PR #41, opened 2022-07-05
Always check assertions

ibbem:
I found this commit in a branch of mine and remember that we talked about this.


pmbittner:
Validation repots the same results and the assertions never fail in the Validation. Curiously, the number of DiffTrees that were filtered because they were empty dropped from 50116 to 25058. This might indicate another bug.

pmbittner (approved):

PR #46, opened 2022-09-12
TikZ export of DiffTrees

ibbem:


pmbittner (commented):
Changes look very good. Proceed as discussed.

src/main/java/org/variantsync/diffdetective/diff/difftree/serialize/GraphvizExporter.java
```diff
@@ -0,0 +1,148 @@
...
+     * Replaces all special characters with uninterpreted escape codes.
+     *
+     * The Graphviz parser for strings interprets some characters specially. The result of this
+     * function can be used in Graphviz strings (surrounded by double quotes) resulting in all
+     * characters appearing literally in the output.
+     *
+     * Note that some backends of Graphviz may still interpret some strings specially, most
+     * commonly strings containing HTML tags.
+     */
+    private String escape(List<String> label) {
```
pmbittner, line 141: Please add documentation for the callee. What is `label` supposed to be? Is it a set of lines?

src/main/java/org/variantsync/diffdetective/diff/difftree/serialize/TikzExporter.java
```diff
@@ -0,0 +1,148 @@
...
+        var output = new PrintStream(destination);
+        output.println("\\begin{tikzpicture}");
+
+        // Convert the layout information received by Graphviz to coordinates used by TikZ.
+        try (var graphizOutput =
+                new BufferedReader(
+                    new InputStreamReader(
+                        new GraphvizExporter(format)
+                            .computeGraphvizLayout(
+                                diffTree,
```
pmbittner, line 69: Is this nesting safe? Are all readers closed upon an exception?

src/main/java/org/variantsync/diffdetective/diff/difftree/serialize/GraphvizExporter.java
pmbittner: Also make this static as it should not and does not rely on object state.

pmbittner (changes requested):
Only minor changes. Looks very good!

src/main/java/org/variantsync/diffdetective/diff/difftree/serialize/TikzExporter.java
ibbem: You're wright, some intermediate constructor might throw an exception (e.g. `OutOfMemoryError`) which would result in possible resource leaks.

ibbem:
All requested changes are done now.

pmbittner (approved):

PR #48, opened 2022-09-17
Adopt the naming convention of the ESEC/FSE 2022 paper

ibbem:
This is based on #46 to avoid heavy merge conflicts when merging. So this should only be merged after #46.


pmbittner:
Please use AE instead of BE. In particular, "artifact" instead of "artefact" to be consistent with the paper.

pmbittner:
Please unflattern the previous "patterns.proposed" package. We would like to keep the "proposed" sub-package.

src/main/java/org/variantsync/diffdetective/diff/difftree/DiffNode.java
```diff
@@ -968,17 +968,17 @@ public boolean isElse() {
...
-        return this.nodeType.equals(NodeType.ROOT);
+        return getBeforeParent() == null && getAfterParent() == null;
     }
 
     /**
      * Returns {@link NodeType#isAnnotation()} for this node's {@link DiffNode#nodeType}.
      */
     public boolean isAnnotation() {
-        return this.nodeType.isAnnotation();
+        return this.nodeType.isAnnotation() && !this.isRoot();
```
pmbittner, line 981: Please remove this extra check. The methods `isAnnotation` and `isArtifact` should partition any DiffTree.

src/main/java/org/variantsync/diffdetective/diff/difftree/NodeType.java
```diff
@@ -15,10 +15,7 @@ public enum NodeType {
     ELIF("elif"),
 
     // Artefact types
-    ARTEFACT("code"),
-
-    // Extra type for the root
-    ROOT("ROOT");
+    ARTEFACT("code");
```
pmbittner, line 18: Rename this to artifact? What would be affected?

ibbem: The following node formats use `NodeType.ARTIFACT.name` for output (`toLabel`):
- DebugMiningDiffNodeFormat
- RWCompositePatternNodeFormat
- DebugDiffNodeFormat
- MappingsDiffNodeFormat
- FormulasAndLineNumbersNodeFormat
- TypeDiffNodeFormat

It's also used by `DiffNode.toString` and
`DiffNodeLabelPrettyfier.prettyPrintTypeAndMapping`.

There is one case where `NodeType.ARTIFACT.name` is used for reading data:
`DebugMiningDiffNodeFormat.fromEncodedTypes` (using `NodeType.fromName`).

Given the *debug* in the name `DebugMiningDiffNodeFormat` I changed the value of
`NodeType.ARTIFACT.name` to "artifact".

pmbittner: As long as the rename is consistent for write _and_ read operations (which should be the case as every such operation should rely on these constans) everything should be fine. We will lose the ability to parse old evaluation data but that is not necessary.

src/main/java/org/variantsync/diffdetective/diff/difftree/DiffNode.java
```diff
@@ -110,7 +110,7 @@ public DiffNode(DiffType diffType, NodeType nodeType,
     public static DiffNode createRoot() {
         return new DiffNode(
                 DiffType.NON,
-                NodeType.ROOT,
+                NodeType.IF,
```
ibbem, line 113: This change causes duplicate IDs on the following diff:
```diff
 #ifdef A
  Code
  #endif
  ```
The root node and its only child have the same ID, because both have the following values:
- from line number: `DiffLineNumber(1, 1, 1)`
- NodeType `NodeType.IF`
- DiffType: `DiffType.NON`

Suggestion by @pmbittner : Change the line number of the root node to `DiffLineNumber.Invalid`.

pmbittner: Good catch! We could change the line number `from` to `DiffLineNumber.Invalid` but we have to make sure that this does not cause trouble with the bit arithmetics. I remember that I did not choose `DiffLineNumber.Invalid` for `from` of the root back then but I don't remember the reason. Maybe it was because of the bit arithmetics or something else. Another option could be to keep the DiffType `ROOT` but only use it for ID conversion and disallow its use as a node's nodeType. Then in `getID`, a node could use ROOT if it is the root and its nodeType otherwise.

pmbittner: A negative line number would be less hacky though and would also highlight that a root that was created by `createRoot` is synthetic and not an actual part of the diffed artifact.

ibbem: The `DiffNode.getID` does add 1 to the line number so negative/invalid line numbers don't cause issues in computing the IDs. Furthermore, he `inDiff` line numbers should be unique for all nodes (no two nodes can be parsed from a single line). Thus, all `DiffNode.getID` should be unique again.

pmbittner: Ok! So `inDiff` should indeed be `DiffLineNumber.Invalid()`. Would you write a unit test to assert that the IDs of an example DiffTree are always unique?

ibbem: A test for this follows in the parser refactoring pull request.

PR #53, opened 2022-10-09
Various refactorings mainly focused on the LineGraph exporter

ibbem:
I'm not sure if the the last commit `9e37848a51f67fbddc20350ea3b279ea7c435b4b` breaks anything, because, to me, the intent of keeping all line graphs in memory was unclear.

Notable behavioral changes:
- `EditClassValidationTask` doesn't add export options to the metadata anymore because it doesn't actually export anything.
- Bug fix: The tree preprocessing in `MiningTask` was run twice:
  First for the line graph export and second directly by the `call` method.


src/main/java/org/variantsync/diffdetective/mining/MiningTask.java
```diff
@@ -48,48 +54,49 @@ public AnalysisResult call() throws Exception {
...
+             * (e.g., match more than one edit class) and export them to the destination
+             * determined by the AnalysisStrategy.
              */
+            int numDiffTrees = 0;
             final CommitDiff commitDiff = commitDiffResult.diff().get();
-            final StringBuilder lineGraph = new StringBuilder();
-            miningResult.append(LineGraphExport.toLineGraphFormat(commitDiff, lineGraph, options.exportOptions()));
-            options.analysisStrategy().onCommit(commitDiff, lineGraph.toString());
-            options.exportOptions().treeFilter().resetExplanations();
+            try (var lineGraphDestination = options.analysisStrategy().onCommit(commitDiff)) {
```
pmbittner, line 63: It's not intuitive to me that an `AnalysisStrategy` always returns a stream. I see that, currently, these strategies are only used to export to file or serve as a dummy. So maybe we should rename `AnalysisStrategy` to `VariationDiffExportStrategy`?

src/test/java/LineGraphTest.java
```diff
@@ -52,8 +50,23 @@ public void idempotentReadWrite() throws IOException {
...
+                    var expectedFile = Files.newBufferedReader(testFile);
+                    var actualFile = Files.newBufferedReader(actualPath);
+            ) {
+                if (!IOUtils.contentEqualsIgnoreEOL(expectedFile, actualFile)) {
+                    Assert.fail("The file " + testFile + " couldn't be exported or imported without modifications");
+                } else {
+                    // Only keep output file on errors
+                    Files.delete(actualPath);
+                }
+            }
```
pmbittner, line 69: Looks good. Previously, this test case never wrote to disk though and now we have several extra file system interactions. But I guess that is fine and we should not complicate the test to avoid that.

src/main/java/org/variantsync/diffdetective/diff/difftree/serialize/LineGraphExport.java
```diff
@@ -91,52 +76,59 @@ public static AnalysisResult toLineGraphFormat(final CommitDiff commitDiff, fina
...
+                options.onError().accept(patchDiff, e);
+                break;
             }
         }
 
         result.exportedCommits = 1;
-        result.filterHits = new ExplainedFilterSummary(options.treeFilter());
 
         return result;
     }
```
pmbittner, line 94: Wow this became so much easier! But I guess you forgot to compose the result of the `toLineGraphFormat` call with the `AnalysisResult result`? Previously all the debug data was collected in the result but now it isn't because the result is ignored, right?

src/main/java/org/variantsync/diffdetective/analysis/strategies/CompositeAnalysisStrategy.java
```diff
@@ -26,18 +27,27 @@ public CompositeAnalysisStrategy(final AnalysisStrategy... strategies) {
...
+        var it = strategies.iterator();
+
+        if (!it.hasNext()) {
+            return OutputStream.nullOutputStream();
         }
+
+        OutputStream destination = it.next().onCommit(commit);
+        while (it.hasNext()) {
+            destination = new TeeOutputStream(destination, it.next().onCommit(commit));
+        }
```
pmbittner, line 48: nice!

src/main/java/org/variantsync/diffdetective/analysis/strategies/AnalyzeAndExportIncrementally.java
```diff
@@ -44,42 +48,40 @@ public AnalyzeAndExportIncrementally() {
...
-
-        if (collectedCommits >= commitsToExportAtOnce) {
-            exportAppend(outputPath, nextChunkToExport.toString());
-            nextChunkToExport = new StringBuilder();
+        if (collectedCommits > commitsToExportAtOnce) {
+            flush();
             collectedCommits = 0;
         }
+
+        return new CloseShieldOutputStream(lineGraphDestination);
```
pmbittner, line 68: What's this stream? Can you document why the usage of such a specific stream is ncessary here?

src/main/java/org/variantsync/diffdetective/analysis/strategies/AnalyzeAllThenExport.java
```diff
@@ -1,34 +1,11 @@
...
-
-    @Override
-    public void end() {
-        final String lineGraph = waitForAll.toString();
-//        Logger.info("Writing file {}", outputPath);
-        IO.tryWrite(outputPath, lineGraph);
+public class AnalyzeAllThenExport extends AnalyzeAndExportIncrementally {
+    public AnalyzeAllThenExport() {
+        super(Integer.MAX_VALUE);
     }
```
pmbittner, line 10: lol ok

src/main/java/org/variantsync/diffdetective/analysis/strategies/AnalyzeAllAndExport.java
```diff
@@ -0,0 +1,48 @@
...
+import org.apache.commons.io.output.CloseShieldOutputStream;
+import org.tinylog.Logger;
+import org.variantsync.diffdetective.datasets.Repository;
+import org.variantsync.diffdetective.diff.CommitDiff;
+import org.variantsync.diffdetective.util.IO;
+
+/**
+ * Collects all linegraph representations generated by an analysis and exports them at the end.
+ * @author Benjamin Moosherr
+ */
```
pmbittner, line 16: From this documentation, the differrence to `AnalyzeAllThenExport` is not clear to me. It seems as if both are the same?

pmbittner: Where is this new class needed?

src/test/java/MarlinDebug.java
```diff
@@ -124,9 +121,6 @@ public static void testCommit(final RepoInspection repoInspection, final String
                 Logger.info("  End processing {}", patch);
             }
         }
-
-        StringBuilder bigB = new StringBuilder();
-        LineGraphExport.toLineGraphFormat(commitDiff, bigB, DiffTreeMiner.MiningExportOptions(repoInspection.repo));
```
pmbittner, line 129: This might have been here to test the filtering and transforming of DiffTrees but this class is deprecated anyway and we should probably delete it or at least flag it as `@Deprecated`.

pmbittner (changes requested):
Only minor comments. It is not easy to see through all the changes at once but they make a good impression and seem to simplify a lot. Greatly appreciated.

src/test/java/LineGraphTest.java
ibbem: I didn't do this because its simpler (it's actually pretty equally simple to keep the line graph in memory using `ByteArrayOutputStream`), but because it has the following two benefits:
- Reduces memory usage: I care at least a little bit about the memory footprint (in this case at least 128K) and let the OS decide whether its worth the effort to write that file on disk.
- More important: It simplifies testing by letting me diff the resulting files if a test case fails.
- There is a third reason for this change which doesn't apply here, but to the same pattern in #54: If I change the code in a way that the test case has to be modified I can easily replace the expected output file by the actual output file after careful inspection.

src/main/java/org/variantsync/diffdetective/analysis/strategies/AnalyzeAndExportIncrementally.java
ibbem: The contract between `AnalysisStrategy.onCommit` and its caller states (in the comment of `AnalysisStrategy.onCommit`) that the returned `OutputStream` is closed after the commit has been processed. However, this strategy wants to reuse the same `OutputStream` for multiple Commits, therefore it has to prevent that the caller closes the stream. That's exactly what `CloseShieldOutputStream` does: It overrides the close method, which does nothing, and forwards all other calls to the `lineGraphDestination`.

I add a comment soon.

src/main/java/org/variantsync/diffdetective/analysis/strategies/AnalyzeAllAndExport.java
ibbem: They differ in their flushing behavior: `AnalyzeAllThenExport` keeps all line graphs in memory until the analysis for all commits has finished and then writes it to disk. In contrast `AnalyzeAllAndExport` writes everything directly to disk (and flushes after every commit).

This is also an easy way to save memory, so I changed the `AnalysisStrategy` in `DiffTreeMiner` from `AnalyzeAllThenExport` to `AnalyzeAllAndExport`. I actually forgot to ask that last week: Is this change even desired? I ask, because `AnalyzeAllThenExport` seemed to be written deliberately to export the line graphs only if the analysis finished successfully.

This comment is actually wrong :sweat_smile:. Will change that soon.

src/test/java/MarlinDebug.java
ibbem: So how did it test that? Was there manual interaction involved? Or did it only check if the filtering or transformation crash?
This would be the second test class (besides `LinuxParsingTest`) annotated with `@Deprecated`.

src/main/java/org/variantsync/diffdetective/diff/difftree/serialize/LineGraphExport.java
ibbem: The overload that is called in that case does return `void` because it passes the `AnalysisResult` as parameter, but I don't remember why I did that.

I will change that too.

src/main/java/org/variantsync/diffdetective/mining/MiningTask.java
ibbem: Well, it doesn't have to return a useful `OutputStream` (see `NullStrategy` for example). Nonetheless, this rename makes sense given the current usage pattern. I will change that in a future PR (as soon as this is merged). Should I open an issue for this (to use the issue tracker as TODO list)?

src/main/java/org/variantsync/diffdetective/analysis/strategies/AnalyzeAllAndExport.java
pmbittner: The goal was to reduce IO operations in favor of a potentially bigger main memory footprint. Otherwise we have an IO operation after each of millions of commits. AnalyzeAllThenExport thus exported commit linegraphs in batches of 1000 commits.

pmbittner: Both strategies are useful. Please keep the old one in DiffTreeMiner for now and refactor your comment here into documentation of the strategies.

src/test/java/MarlinDebug.java
pmbittner: It was dome manually if I remember correctly. Dont take this code too serious. :D

src/test/java/LineGraphTest.java
pmbittner: ok

src/main/java/org/variantsync/diffdetective/analysis/strategies/AnalyzeAllAndExport.java
ibbem: The `DiffTreeMiner` is now using `AnalyzeAllThenExport` again and the comment of `AnalyzeAllAndExport` is referencing `AnalyzeAndExportIncrementally` for clarity.

pmbittner (approved):

PR #54, opened 2022-10-16
Refactor the DiffTreeParser and upgrade to JUnit5

ibbem:
The code for parsing `DiffTree`s is now smaller (considering all the code removed in other classes). This is mostly due to the introduction of the `LogicalLine` which did make most of the other multi line macro parsing obsolete. This change also reflects the actual C11 standard more closely in nomenclature and how it should actually work.
This refactoring does introduce breaking changes, but this should be limited to edge cases which are now parsed conforming to the C11 standard or rejected if they are represent an invalid program (before or after the change). There are also some edge cases (which are hopefully very rare in practice) which are not handled correctly by this parser (the old code didn't either). Examples include:
```diff
 # \
    ifdef A
 #endif

 # /*
   Comment
   */ ifdef A
 #endif

/*
  #ifdef Comments are removed before the preprocessor is run.
*/
```
Most of the time such programs are rejected by the parser, but there are cases where this isn't the case, for example:
```diff
 /*
   #ifdef A
      This code is considered as annotated by Feature A, although it isn't according to C11.
   #endif
 */
```

For easier testing I upgraded JUnit4 to JUnit5. This enables us to use features like parameterized tests which I used to replace loops over different test cases. This makes such test cases independent (if some fail, the others are still run), prints the specific test case which actually caused a test failure and allows the selection of one such specific test case for the next run, for example to debug it.
Furthermore the `ExportTest` (using Graphviz) is now disabled if Graphviz can't be executed.

During testing I found a bug in the `GraphvizExporter`: The escaping didn't work correctly (java threw an exception if `GraphvizExporter.escape` encountered a backshlash). This is fixed by 478c8b493353ed8cae64aaee56049b9a302cded8.


pmbittner:
I guess this PR builds on top of other PRs? Which PRs should we merge first?

ibbem:
No #53 and this PR are independent. There are a some merge conflicts in imports and tests but nothing too complicated. If you don't want to deal with these conflicts you can wait until #53 is merged so I can rebase this PR.

pmbittner:
Ok then I will review this PR soon.

src/main/java/org/variantsync/diffdetective/diff/difftree/DiffType.java
```diff
@@ -2,6 +2,8 @@
 
 import java.util.function.Consumer;
 
+import org.apache.commons.lang3.function.FailableConsumer;
```
pmbittner, line 3: We also have interfaces for this in `Functjonal`. They are called `FragileSupplier`... I did not know that Apache Commons also had these. It would probably be good to be consistent and always use the same interfaces? The Functjonal-Interfaces also have support within Functjonal, for example for the `Result` type. What is your opinion?

src/main/java/org/variantsync/diffdetective/diff/difftree/DiffType.java
```diff
@@ -53,12 +55,18 @@ public void matchBeforeAfter(final Runnable ifExistsBefore, final Runnable ifExi
...
      * @param ifExistsAfter Argument that is valid if the edit did not remove.
      * @param task Task to run with all given arguments that are valid w.r.t. to this DiffType's lifetime.
+     * @throws E iff {@code task} throws {@code E}
      */
-    public <T> void matchBeforeAfter(final T ifExistsBefore, final T ifExistsAfter, final Consumer<T> task) {
+    public <T, E extends Throwable> void matchBeforeAfter(
+        final T ifExistsBefore,
+        final T ifExistsAfter,
+        final FailableConsumer<T, E> task
+    ) throws E {
```
pmbittner, line 68: This cannot be called anymore with a default consumer, right? Would be nice to be able to run this method with a non-failing consumer but as long as we do not need it right now, we do not have to add it.

src/main/java/org/variantsync/diffdetective/diff/difftree/parse/DiffTreeParser.java
```diff
@@ -102,35 +96,24 @@ public static DiffResult<DiffTree> createDiffTree(
...
-                case Error: {
-                    isMLMacro.onError(errorPropagation);
-                    return error.get();
-                }
-                // line is not a mult-line macro so keep going (break the switch statement).
-                case NotMyDuty: break;
+            } catch (IllFormedAnnotationException e) {
+                throw new DiffParseException(e.getType(), lineNo);
             }
+            // line is not a mult-line macro so keep going (break the switch statement).
```
pmbittner: The comment should be adapted because `(break the switch statement)` does not apply anymore.

src/main/java/org/variantsync/diffdetective/diff/difftree/parse/MultiLineMacroParser.java
```diff
@@ -84,13 +81,13 @@ ParseResult consume(
...
                     if (beforeMLMacro != null) {
-                        return ParseResult.ERROR(MLMACRO_WITHIN_MLMACRO, "Found definition of multiline macro within multiline macro at line " + line + "!");
+                        throw new DiffParseException(DiffError.MLMACRO_WITHIN_MLMACRO, lineNo);
                     }
                     beforeMLMacro = new MultilineMacro(line, diffType, lineNo, beforeStack.peek(), afterStack.peek());
                 }
                 if (!isRem) {
                     if (afterMLMacro != null) {
-                        return ParseResult.ERROR(MLMACRO_WITHIN_MLMACRO, "Found definition of multiline macro within multiline macro at line " + line + "!");
+                        throw new DiffParseException(DiffError.MLMACRO_WITHIN_MLMACRO, lineNo);
```
pmbittner: Very nice that you keep the line number here! :)

src/main/java/org/variantsync/diffdetective/diff/difftree/parse/MultiLineMacroParser.java
```diff
@@ -140,15 +135,8 @@ ParseResult consume(
...
-                    if (pushResult.isError()) {
-                        return pushResult;
-                    }
-
-                    pushResult = DiffTreeParser.pushNodeToStack(mlNode, afterStack, afterMLMacro.getLineFrom());
-                    if (pushResult.isError()) {
-                        return pushResult;
-                    }
+                    DiffTreeParser.pushNodeToStack(mlNode, beforeStack, beforeMLMacro.getLineFrom());
+                    DiffTreeParser.pushNodeToStack(mlNode, afterStack, afterMLMacro.getLineFrom());
```
pmbittner: And suddenly we have `do` notation :D

src/main/java/org/variantsync/diffdetective/diff/result/CommitDiffResult.java
```diff
@@ -19,9 +20,10 @@ public record CommitDiffResult(Optional<CommitDiff> diff, List<DiffError> errors
      * @return A failure result that was caused by the given error.
      */
     public static CommitDiffResult Failure(DiffError error, String message) {
+        Logger.debug("[CommitDiffResult::Failure] {}", message);
```
pmbittner, line 23: The `"[CommitDiffResult::Failure]"` should be inserted automatically by Log4J so it is redundant here, isn't it?

src/main/java/org/variantsync/diffdetective/diff/difftree/DiffType.java
```diff
@@ -93,15 +92,17 @@ public DiffType inverse() {
...
+     * @return The type of edit of <code>line</code> or null if its an invalid diff line.
      */
-    public static DiffType ofDiffLine(String line) {
+    public static DiffType ofDiffLine(String line) throws DiffParseException {
         if (line.startsWith(ADD.symbol)) {
             return ADD;
         } else if (line.startsWith(REM.symbol)) {
             return REM;
-        } else {
+        } else if (line.startsWith(NON.symbol)) {
```
pmbittner, line 102: Good catch. This invalidates a handy "bug": The parsing of variation trees. When interpreting lines not starting with + or - as unchanged, we can also parse normal code. Doing so, we can parse C code to get variation trees instead of variation diffs. Admittedly, your fix here is correct here but previously this "bug" had a nice side-effect and some of my use-cases outside of DiffDetective might break.

As a better solution to exploiting this behaviour, is to keep your fix here, but create a `VariationTreeParser` next to the `DiffTreeParser` that just forwards the parsing of a file or string to the DiffTreeParser after prepending a space ` ` to every line.

src/main/java/org/variantsync/diffdetective/diff/difftree/parse/DiffTreeParser.java
```diff
@@ -12,21 +12,64 @@
...
+    /* Settings */
+    final boolean collapseMultipleCodeLines;
+    final boolean ignoreEmptyLines;
+    final CPPAnnotationParser annotationParser;
+
+    /* State */
+    private final Stack<DiffNode> beforeStack = new Stack<>();
+    private final Stack<DiffNode> afterStack = new Stack<>();
+
+    private DiffNode lastArtifact = null;
```
pmbittner, line 69: Please document what this field is used for.

src/main/java/org/variantsync/diffdetective/diff/difftree/parse/LogicalLine.java
```diff
@@ -0,0 +1,80 @@
+package org.variantsync.diffdetective.diff.difftree.parse;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.variantsync.diffdetective.diff.DiffLineNumber;
+import org.variantsync.diffdetective.util.Assert;
+
+/**
+ * A Logical line consisting of multiple physical lines joined by line continuations.
```
pmbittner, line 10: Please add `@author` and elaborate on the meaning of `line`. The word `line` is more general than lines within text-based files as we inspect.

src/main/java/org/variantsync/diffdetective/diff/difftree/DiffType.java
pmbittner: https://xkcd.com/1172/

src/main/java/org/variantsync/diffdetective/diff/difftree/serialize/edgeformat/ChildOrderEdgeFormat.java
```diff
@@ -0,0 +1,12 @@
...
+import org.variantsync.diffdetective.diff.difftree.serialize.StyledEdge;
+
+public class ChildOrderEdgeFormat extends EdgeLabelFormat {
+    @Override
+    public String labelOf(StyledEdge edge) {
+        int i = edge.from().indexOfChild(edge.to());
+        int j = edge.to().indexOfChild(edge.from());
+        return String.format(";%d", i < 0 ? j : i);
+    }
+}
```
pmbittner, line 12: Please add documentation on author and how labels are formatted.

src/test/java/ExportTest.java
```diff
@@ -48,17 +53,35 @@ public static boolean isGraphvizInstalled() throws InterruptedException {
...
-        TestUtils.assertEqualToFile(Path.of("src/test/resources/serialize/expected.tex"), tikzOutput.toString());
+        try (
+                var expectedFile = Files.newBufferedReader(expectedPath);
+                var actualFile = Files.newBufferedReader(actualPath);
+        ) {
+            if (!IOUtils.contentEqualsIgnoreEOL(expectedFile, actualFile)) {
+                fail("The DiffTree in file " + testCasePath + " didn't parse correctly. "
+                    + "Expected the content of " + expectedPath + " but got the content of " + actualPath + ". ");
+            } else {
+                // Keep output file for debugging on errors
```
pmbittner, line 82: Comments is a bit confusing as is right now because it says "keep file" directly in front of "Files.delete". Please adapt to "delete files of successful tests" or something like that.

pmbittner (changes requested):
The changes look very good. Thank you! I appreciate the simplifications in the parser and in the tests by upgrading to JUnit 5. The changes in the parser almost replace the parser completely, so it was not possible for me to check on it just from scrolling through big git diffs. The changes made a good impression though and having the test cases should give us more certainty in the correctness of the new parser than the old one. I found some documentation issues. I will also see how the new parser behaves within the ESEC/FSE22 validation soon, just to have a comparison.

src/main/java/org/variantsync/diffdetective/diff/difftree/DiffType.java
ibbem: It cannot be called if you have an explicit instance of `Consumer`, for example using the old anonymous class style: 
```java
diffType.matchBeforeAfter(1, 2, new Consumer() {
  @Override
  void accept(Integer a) {
    // Do something with a
  }
}
```
If you instead call it with a lambda it works just fine (if the lambda does not throw any exceptions the type `E` is resolved to `RuntimeException` and `FailableConsumer<T, RuntimeException>` and `Consumer<T>` are functionally equivalent).

However, when adding an overload for `Consumer<T>` the overload resolutions fails when using a lambda, which would be rather unfortunate.

src/main/java/org/variantsync/diffdetective/diff/difftree/DiffType.java
ibbem: I checked their source code: These interfaces are almost identical. The only difference is that Functjonal uses `Exception` as bound and Apache Commons Lang3 uses `Throwable`, which is more general.

I propose to
- remove the `functions` package from Functjonal
- add a Apache Commons Lang3 dependency to Functjonal
- use `Failable*` instead of `Fragile*` everywhere

Pro:
- Java has a nominal type system, so two identical interfaces are not compatible. That will probably result in pain as soon as some dependency of DiffDetective uses Apache Commons Lang3.
- Apache Commons Lang3 already has failable alternatives for most (probably all, but I didn't check that) of the standard functional interfaces in `java.util.function`.
- Apache Commons Lang3 also provides some useful helpers, for example converting a `FailableX` to an `X` by rethrowing all exceptions as `RuntimeException`. It also provides a `FailableStream`.

Cons:
- Adding a Apache Commons Lang3 dependency to Functjonal.
- This is a breaking change for users of Functjonal. I don't know how many users Functjonal has, so I can't estimating how much impact that would have.

src/main/java/org/variantsync/diffdetective/analysis/AnalysisResult.java
```diff
@@ -239,9 +239,12 @@ public static AnalysisResult importFrom(final Path p, final Map<String, BiConsum
                         } else if (key.startsWith(ExplainedFilterSummary.FILTERED_MESSAGE_BEGIN)) {
                             filterHitsLines.add(line);
                         } else if (key.startsWith(ERROR_BEGIN)) {
-                            DiffError e = new DiffError(key.substring(ERROR_BEGIN.length(), key.length() - ERROR_END.length()));
+                            var e = DiffError.fromMessage(key.substring(ERROR_BEGIN.length(), key.length() - ERROR_END.length()));
+                            if (e.isEmpty()) {
+                                throw new RuntimeException("Invalid error id");
```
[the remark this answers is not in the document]
ibbem: This message now includes the error key and the path of the file containing the error.

src/main/java/org/variantsync/diffdetective/diff/difftree/parse/DiffTreeParser.java
ibbem: I just removed the part in the brackets. This comment is also removed in a later commit, so I mark this as resolved.

src/main/java/org/variantsync/diffdetective/diff/result/CommitDiffResult.java
ibbem: Good catch. Yes you're right. I removed this prefix now.

src/main/java/org/variantsync/diffdetective/diff/difftree/parse/DiffTreeParser.java
ibbem: I added documentation for all fields. (I remember that I wanted to do that anyway, but forgot it somehow.)

src/main/java/org/variantsync/diffdetective/diff/difftree/parse/LogicalLine.java
ibbem: Good catch. I didn't add `@author` anywhere. This tag is now included in every file I created.
Whats your policy for changing `@author`? For example the `DiffTreeParser` doesn't have an `@author` annotation. Should I add one. If yes, who should be mentioned?

I didn't quite get the second part. What other meanings of line are there?
The logical and physical part is now explained by referencing the C11 standard and providing a regex for more rigorous explanation. What else should I mention?

src/main/java/org/variantsync/diffdetective/diff/difftree/serialize/edgeformat/ChildOrderEdgeFormat.java
ibbem: Documentation and `@author` added.

src/test/java/ExportTest.java
ibbem: You're right. There are now two comments explaining the action in both branches.

src/main/java/org/variantsync/diffdetective/diff/result/DiffError.java
```diff
@@ -1,61 +1,81 @@
...
+            .filter(m -> m.getMessage().equals(message))
+            .findFirst();
+    }
 
     @Override
     public String toString() {
-        return id;
+        return message;
     }
 }
```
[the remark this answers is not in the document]
ibbem: In my opinion all errors produces by the code in DiffDetective should be added to such an enum. If there is someone who wants to handle some errors, but not all, he can differentiate between these errors without relying on a string which is probably not really public API. The compiler can also provide errors if the error name is changed or doesn't exist any more.
If the users of DiffDetective as a library should be able to add new error codes, it might be a better idea to introduce an interface for such error codes and let the enum implement this interface. Although I currently don't know of a use case where users of DiffDetective should be able to pass custom errors into DiffDetective.

src/main/java/org/variantsync/diffdetective/diff/GitDiffer.java
```diff
@@ -308,18 +307,26 @@ private static CommitDiffResult getPatchDiffs(
...
+                            file,
+                            parseOptions
                         );
 
-                patchDiff.ifSuccess(commitDiff::addPatchDiff);
-                patchDiff.ifFailure(errors::add);
+                    commitDiff.addPatchDiff(patchDiff);
+                } catch (IOException e) {
+                    Logger.debug(e, "Could not obtain full diff of file " + filename + " before commit " + parentCommit + "!");
+                    errors.add(DiffError.COULD_NOT_OBTAIN_FULLDIFF);
```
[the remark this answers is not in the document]
ibbem: Compared to before, I do think this is better. Most error constructors didn't really store the error message. They just logged it and forgot about it. When I first read these methods I was really confused why I have to pass a message into a function which just logs it for me.
The main thing I like to change in the future is to provide the filename and the commit hash inside of the error, so that the caller can decide what to do with them. This should removed the need for log & throw.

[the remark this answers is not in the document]
ibbem: Note there still are such "just log the message" smart constructors. See `CommitDiffResult` below.

ibbem:
The requested changes are now finished. There are also some new modifications of javadoc which I missed initially. The most notable change is the new commit for parsing variation trees from source files.

src/main/java/org/variantsync/diffdetective/diff/result/DiffError.java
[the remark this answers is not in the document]
pmbittner: I agree. The current solution is probably the best and definitely a good one for our use cases so far.

src/main/java/org/variantsync/diffdetective/diff/GitDiffer.java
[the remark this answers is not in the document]
pmbittner: Ok

src/main/java/org/variantsync/diffdetective/diff/difftree/parse/DiffTreeParser.java
```diff
@@ -153,7 +165,51 @@ public static DiffTree createDiffTree(
...
+            ignoreEmptyLines,
+            annotationParser
+        ).parse(() -> {
+            String line = file.readLine();
+            if (line == null) {
+                return null;
+            } else {
+                return new DiffLine(DiffType.NON, line);
+            }
+        });
```
pmbittner, line 212: The changes look very good, thank you. It would be great though if this method should throw an error when the diff contains changed lines. We can only parse variation trees from source code files and not from diffs. So even though this is a workaround anyway, it should not accept changed lines in a diff.

pmbittner (changes requested):

src/main/java/org/variantsync/diffdetective/diff/difftree/DiffType.java
pmbittner: Thanks for the change. Comments below.

src/main/java/org/variantsync/diffdetective/diff/difftree/parse/LogicalLine.java
pmbittner: Ok thank you. I guess I forgot to add the author tag at some places too. Just leave these places for now.

I was referring to something like "line of text" or "line of code" because a just a line is also something that you can draw with a pencil. :D

pmbittner:
Thank you for the changes. Only a few comments are left to discuss / resolve. :)

src/main/java/org/variantsync/diffdetective/diff/difftree/parse/DiffTreeParser.java
ibbem: This method (`createVariationTree`) doesn't receive a diff, but a source file, so it will never parse changed lines. If you provide it a diff, then it will interpret the changed lines as unchanged source lines which start with a plus or minus by chance.
Do I understand correctly that you want to check if any line starts with a plus or minus to prevent accidentally misusing `createVariationTree` for parsing diffs? This heuristic would also reject some well formed source code (although in most programming languages this would be very badly formatted code).

src/main/java/org/variantsync/diffdetective/diff/difftree/parse/LogicalLine.java
ibbem: Oh, now I get it. It makes sense that not everybody associates logical and physical lines with source code :D.
I added that physical lines are
> of a text file

to the summary of the class.

src/main/java/org/variantsync/diffdetective/diff/difftree/parse/DiffTreeParser.java
pmbittner: Yes, preventing (accidental) misuse was my intent but I see your point that the proposed heuristic would also prevent some legal input. What about keeping the code as is but issuing a warning when a + or - is found at the beginning of a line here?

ibbem:
Note: My last force push was to rebase this branch onto develop to resolve the conflicts.

src/main/java/org/variantsync/diffdetective/diff/difftree/parse/DiffTreeParser.java
ibbem: This is probably a good idea as this will probably not often appear in practice.

pmbittner (approved):

PR #56, opened 2022-10-24
Updates commit count in docs/datasets.md.

SebastianWilke:
Implicitly verifies that every `Repository URL` still works.
The only project for which commit count went down was [Marlin](https://github.com/MarlinFirmware/Marlin) for some reason.


pmbittner:
Hi @SebastianWilke ,

thanks for your pull-request and checking that the URLs are still valid. What was your motivation for updating the commit counts? All these repositories get commits frequently, so the expected commit counts is always out of date, and thus impossible to keep up to date anyway. That's why we referred to this entry as just `Estimated number of commits`. Its purpose is to inform humans on expected repo size but the numbers should not be used in a computation.

SebastianWilke:
Hi @pmbittner, I'm currently writing my bachelors thesis under the supervision of A. Schultheiß. I've used the list you compiled in order to check how many of the projects use the CMake build system (something relevant for a claim I make in my work). Contrary to what you say, I figured the commit count would be used as part of some computation and thought this might be a way to give back for creating the list. I am not disappointed in any way if the changes get not approved.

pmbittner:
Ok thank you. I am happy that the list was helpful. :) I guess the commit count is parsed in the code base and stored within a `DatasetDescription` but it is never used. Could you redirect the pull request to the `develop` branch instead of `main`? Then I will merge.

SebastianWilke:
@pmbittner Done.

PR #59, opened 2022-11-06
Implemention of variation trees

ibbem:
Implements variation trees and generalizes some algorithms by using projections of variation diffs.

There is some more refactoring potential here (for example `DiffTreeSource` could use two `VariationTreeSource`s) and there are very few convenience methods in `VariationTree` in comparison to `DiffTree` but this can be fixed later, when needed.


pmbittner:
Ok, we should merge the PR in Functjonal first.

src/main/java/org/variantsync/diffdetective/variationtree/VariationNode.java
```diff
@@ -0,0 +1,583 @@
...
+
+    /**
+     * Returns the end line number of the {@link getLabelLines label} of this node in the
+     * corresponding source code.
+     *
+     * <p>The line number is exclusive (i.e., it points 1 behind the last included line).
+     *
+     * @see setToLine
+     */
+    public abstract int getToLine();
```
pmbittner, line 96: For consistency with `DiffLineNumber` and `Lines`, this should probably be called `getToLineExclusive`. It's a bit verbose but I hope that this helps to avoid some indexing bugs for users right away. Please rename this method and all the other methods acting on line numbers in this class where appropriate.

src/main/java/org/variantsync/diffdetective/variationtree/VariationNode.java
```diff
@@ -0,0 +1,583 @@
...
+    /**
+     * Computes the number of {@link isAnnotation annotations} of the path from the root node to
+     * this node.
+     */
+    public int getAnnotationDepth() {
+        if (isRoot()) {
+            return 0;
+        }
+
+        if (isIf()) {
```
pmbittner, line 188: or Elif or Else? Where is this method used?

src/main/java/org/variantsync/diffdetective/variationtree/VariationNode.java
```diff
@@ -0,0 +1,583 @@
...
+        return getParent().getIfNode();
+    }
+
+    /**
+     * Returns {@code true} iff this node is the parent of the given node.
+     *
+     * @see getChildren
+     */
+    public boolean isChild(VariationNode<T> child) {
+        return child.getParent().isSameAs(child);
```
pmbittner, line 214: `child.getParent().isSameAs(this)`?

src/main/java/org/variantsync/diffdetective/variationtree/VariationNode.java
```diff
@@ -0,0 +1,583 @@
...
+    /**
+     * Adds a child before the given index to the children list of this node and sets its parent to
+     * this node.
+     *
+     * <p>When calling {@link indexOfChild} with {@code child} the returned index will be
+     * {@code index} as long as the children list isn't modified.
+     *
+     * @see addChildren
+     */
+    public abstract void addChild(final VariationNode<T> child, int index);
```
pmbittner, line 247: insertChild?

src/main/java/org/variantsync/diffdetective/variationtree/VariationNode.java
```diff
@@ -0,0 +1,583 @@
...
+
+    /**
+     * Adds this subtree below the given parent.
+     * Inverse of {@link drop}.
+     *
+     * @see addChild
+     */
+    public void addBelow(final VariationNode<T> parent) {
+        if (parent != null) {
+            parent.addChild(this);
```
pmbittner, line 295: We also have to check that `this` does not have have a parent already.

src/main/java/org/variantsync/diffdetective/variationtree/VariationNode.java
```diff
@@ -0,0 +1,583 @@
...
+        clauses.add(getDirectFeatureMapping());
+        return clauses;
+    }
+
+    /**
+     * Returns the presence condition of this node.
+     * See Equation (2) in <a href="https://doi.org/10.1145/3540250.3549108">our paper</a>.
+     */
+    public Node getPresenceCondition() {
+        return new And(getPresenceConditionClauses());
```
pmbittner, line 437: Apply same optimization as for `getFeatureMapping()`? If the number if clauses is 1, we can just return that single clause without wrapping it within an And node.

src/main/java/org/variantsync/diffdetective/variationtree/VariationNode.java
```diff
@@ -0,0 +1,583 @@
...
+    }
+
+    /**
+     * Traverses all nodes in this subtree with in preorder.
```
pmbittner, line 441: remove `with`?

src/main/java/org/variantsync/diffdetective/variationtree/VariationNode.java
```diff
@@ -0,0 +1,583 @@
...
+     */
+    public void forAll(Consumer<VariationNode<T>> action) {
```
pmbittner, line 443: Name this method by its traversal order? `forAllPreorder`?

src/main/java/org/variantsync/diffdetective/variationtree/VariationNode.java
```diff
@@ -0,0 +1,583 @@
...
+
+    /**
+     * Returns the first {@code if} node in the path from this node upwards to the root.
+     */
+    public VariationNode<T> getIfNode() {
+        if (isIf()) {
+            return this;
+        }
+        if (isRoot()) {
+            return null;
```
pmbittner, line 203: Semantically, this is a dead branch. By our invariant, the root is always an `IF`. We can keep this branch as a security check here but probably its better to throw an Exception instead of returning `null`.

pmbittner (changes requested):
The changes look awesome! Thank you. I still have to review `VariationTree` and `VariationTreeNode` which I will do asap.

src/main/java/org/variantsync/diffdetective/variationtree/VariationTree.java
```diff
@@ -0,0 +1,121 @@
...
+    public static VariationTree fromFile(
+        final BufferedReader input,
+        final VariationTreeSource source,
+        final boolean collapseMultipleCodeLines,
+        final CPPAnnotationParser annotationParser
+    ) throws IOException, DiffParseException {
+        VariationTreeNode tree = DiffTreeParser
+            .createVariationTree(input, collapseMultipleCodeLines, false, annotationParser)
+            .getRoot()
+            .projection(BEFORE)
```
pmbittner, line 111: Leave a note here that BEFORE and AFTER yield equivalent results and so just any of them is fine?

src/main/java/org/variantsync/diffdetective/variationtree/VariationTreeNode.java
```diff
@@ -0,0 +1,364 @@
...
+import org.variantsync.diffdetective.util.fide.FixTrueFalse;
+
+import java.util.*;
+
+/**
+ * A single node in a variation tree.
+ *
+ * <p>A variation tree is a representation of source code in a software product line. Each node in
+ * such a tree is either an {@link isArtifact artifact} or a {@link isAnnotation mapping node}.
+ * Artifacts represent source code which may be reused in multiple variations. In contrast, mappings
```
pmbittner, line 19: variations -> variants

src/main/java/org/variantsync/diffdetective/variationtree/VariationTreeNode.java
```diff
@@ -0,0 +1,364 @@
...
+ * do not contain source code of a specific variant. They store information about which variants
+ * contain the artifacts it has as {@link getChildren children} in the form of a
+ * {@link getFeatureMapping feature mapping}.
+ *
+ * <p>See definition 2.2 of <a href="https://doi.org/10.1145/3540250.3549108">our ESEC/FSE'22
+ * paper</a> for a mathematical formalization or variation trees. In contrast to this formalization,
+ * this implementation optimizes for variation trees obtained from source code annotated with the C
+ * preprocessor syntax. This mostly means that there is the additional {@link NodeType#ELIF} node
+ * type.
```
pmbittner, line 28: Do you need a closing `</p>`?

src/main/java/org/variantsync/diffdetective/variationtree/VariationTreeNode.java
```diff
@@ -0,0 +1,364 @@
...
+ * <p>To compare two variation trees use {@link DiffTree} which uses the aforementioned
+ * {@link DiffNode}s.
+ *
+ * @see VariationTree
+ * @see DiffTree
+ * @see DiffNode
+ * @author Benjamin Moosherr
+ */
+public class VariationTreeNode extends VariationNode<VariationTreeNode> {
+    private static final short ID_OFFSET = 3;
```
pmbittner, line 49: Maybe we should store this somewhere else because currently this is duplicated. The ID_OFFSET depends on the number of values of DiffType and NodeType (it says how many bits are necessary to store each one of them). What about putting this into NodeType?

src/main/java/org/variantsync/diffdetective/variationtree/VariationTreeNode.java
```diff
@@ -0,0 +1,364 @@
...
+
+    /**
+     * The start line number of the {@link label} of this node in the corresponding source code.
+     */
+    private int from = DiffLineNumber.InvalidLineNumber;
+    /**
+     * The end line number of the {@link label} of this node in the corresponding source code.
+     * The line number is exclusive (i.e., it points 1 behind the last included line).
+     */
+    private int to = DiffLineNumber.InvalidLineNumber;
```
pmbittner, line 65: Explicit naming? `toExclusive?`

src/main/java/org/variantsync/diffdetective/variationtree/VariationTreeNode.java
```diff
@@ -0,0 +1,364 @@
...
+    /**
+     * The node type of this node, which determines the type of the represented
+     * element in the diff (e.g., mapping or artifact).
+     */
+    public final NodeType nodeType;
+
+    /**
+     * The start line number of the {@link label} of this node in the corresponding source code.
+     */
+    private int from = DiffLineNumber.InvalidLineNumber;
```
pmbittner, line 60: Explicit naming? `fromInclusive`? Reusing `Lines` here would probably cause more trouble than it helps avoiding ambiguities with offsets.

src/main/java/org/variantsync/diffdetective/variationtree/VariationTreeNode.java
```diff
@@ -0,0 +1,364 @@
...
+     */
+    private final List<VariationTreeNode> childOrder;
+
+    /**
+     * Creates a new node of a variation tree.
+     *
+     * The newly created node is not connected to any other nodes.
+     *
+     * The given label is not copied, so modifications to {@code label} reflect in the label of this
+     * node. It's recommended to remove all references to {@code label} after the node is fully
```
pmbittner, line 99: I do not understand `modifications to label reflect in the label of this node`.

pmbittner: What about saying "takes ownership of the list without copying"?

src/main/java/org/variantsync/diffdetective/variationtree/VariationTreeNode.java
```diff
@@ -0,0 +1,364 @@
...
+     * parsed. This makes it possible to store the unmodifiable result of {@link getLabelLines}
+     * without making a copy.
+     *
+     * @param nodeType the type of this node
+     * @param fromLine the start line number of the code of {@code label}
+     * @param toLine the end line number (exclusive) of the code of {@code label}
+     * @param featureMapping the direct feature mapping of this node, has be nun null iff
```
pmbittner, line 106: nun -> non

src/main/java/org/variantsync/diffdetective/variationtree/VariationTreeNode.java
```diff
@@ -0,0 +1,364 @@
...
+        this.from = fromLine;
+        this.to = toLine;
+        this.label = label;
+        this.featureMapping = featureMapping;
+
+        this.childOrder = new ArrayList<>();
+    }
+
+    /**
+     * Convenience constructor for creating the root of an {@link VariationTree}.
```
pmbittner, line 131: an -> a

src/main/java/org/variantsync/diffdetective/variationtree/VariationTreeNode.java
```diff
@@ -0,0 +1,364 @@
...
+    }
+
+    @Override
+    public NodeType getNodeType() {
+        return nodeType;
+    }
+
+    @Override
+    public List<String> getLabelLines() {
+        return label;
```
pmbittner, line 175: Document that this is a modifiable reference?

src/main/java/org/variantsync/diffdetective/variationtree/VariationTreeNode.java
```diff
@@ -0,0 +1,364 @@
...
+    public void addChild(final VariationNode<VariationTreeNode> child, int index) {
+        child.getBackingNode().setParent(this);
+        childOrder.add(index, child.getBackingNode());
+    }
+
+    @Override
+    public boolean removeChild(final VariationNode<VariationTreeNode> child) {
+        if (isChild(child)) {
+            child.getBackingNode().parent = null;
+            childOrder.remove(child);
```
pmbittner, line 256: `childOrder.remove(child.getBackingNode())`? Why doesn't Java complain here? I guess because the list methods just require `Object`s for some reason. 🙄

src/main/java/org/variantsync/diffdetective/variationtree/VariationTreeNode.java
```diff
@@ -0,0 +1,364 @@
...
+        this.parent = newParent;
+    }
+
+    @Override
+    public List<VariationNode<VariationTreeNode>> getChildren() {
+        return Collections.unmodifiableList(childOrder);
+    }
+
+    @Override
+    public void addChild(final VariationNode<VariationTreeNode> child) {
```
pmbittner, line 241: So to all these methods, we could also pass other nodes that wrap `VariationTreeNode`s. Is this something we want to allow? What about making these methods only accept a value of the backing type (here: VariationTreeNode)? Would this break something with the Projections or would it be a fine restriction?

pmbittner: It would also avoid to invoke `getBackingNode` everywhere which could sometimes easily be forgotten (see below).

src/main/java/org/variantsync/diffdetective/diff/difftree/transform/NaiveMovedArtifactDetection.java
```diff
@@ -34,8 +37,8 @@ public void transform(final DiffTree diffTree) {
                 removed = twin.first();
             }
 
-            final DiffNode afterParent = added.getAfterParent();
-            final DiffNode beforeParent = removed.getBeforeParent();
+            final DiffNode afterParent = added.getParent(AFTER);
+            final DiffNode beforeParent = removed.getParent(BEFORE);
```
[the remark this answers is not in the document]
ibbem: Well, I'm in favor of such changes as long as we know all call sites because it keeps the code cleaner and simpler. Without this change we would have almost three times as much methods in `DiffNode` for literally no advantage on the call sites.
If you're concerned about the time consumption: Using Neovim I developed a workflow for such things which makes this pretty easy and fast.

src/main/java/org/variantsync/diffdetective/diff/difftree/DiffNode.java
```diff
@@ -714,172 +523,51 @@ private List<Node> getFeatureMappingClauses(final Function<DiffNode, DiffNode> p
...
             and.add(negate(ancestor.getDirectFeatureMapping()));
 
             return and;
         } else if (isArtifact()) {
-            return parent.getFeatureMappingClauses(parentOf);
+            return parent.getFeatureMappingClauses(time);
         }
 
         return List.of(getDirectFeatureMapping());
     }
```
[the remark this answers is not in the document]
ibbem: Thanks.

src/main/java/org/variantsync/diffdetective/variationtree/VariationTreeNode.java
```diff
@@ -0,0 +1,364 @@
...
+    @Override
+    public int getID() {
+        // Add one to ensure invalid (negative) line numbers don't cause issues.
+        int lineNumber = 1 + from;
+        Assert.assertTrue((lineNumber << 2*ID_OFFSET) >> 2*ID_OFFSET == lineNumber);
+
+        int id;
+        id = lineNumber;
+        id <<= ID_OFFSET;
+        // This makes `VariationTreeNode.toID` compatible with `DiffNode.toID`
```
pmbittner, line 294: good!

src/main/java/org/variantsync/diffdetective/variationtree/VariationTreeNode.java
```diff
@@ -0,0 +1,364 @@
...
+
+        final int nodeTypeOrdinal = id & lowestBitsMask;
+        Assert.assertTrue(((id >> ID_OFFSET) & lowestBitsMask) == DiffType.NON.ordinal());
+        final int from = (id >> (2*ID_OFFSET)) - 1;
+
+        return new VariationTreeNode(
+                NodeType.values()[nodeTypeOrdinal],
+                from,
+                DiffLineNumber.InvalidLineNumber,
+                null,
```
pmbittner, line 320: Use true if we have an annotation node? Maybe its best to keep null here though as a "reminder" that this node is still incomplete.

pmbittner (changes requested):

src/main/java/org/variantsync/diffdetective/diff/difftree/DiffNode.java
```diff
@@ -1112,6 +764,204 @@ public String toTextDiff() {
...
+        @Override
+        public int getID() {
+            return getBackingNode().getID();
+        }
+
+        @Override
+        public boolean isSameAs(VariationNode<DiffNode> other) {
+            if (other != null && getClass() == other.getClass()) {
+                Projection otherProjection = (Projection) other;
+                return time.equals(otherProjection.time) && getBackingNode() == otherProjection.getBackingNode();
```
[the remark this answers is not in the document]
ibbem: Well it's documented that you should use `isSameAs` instead of `==` for identity tests.  In this case it's an implementation detail that `DiffNode.Projection` guarantees that this works and therefore it's documented in `DiffNode.Projection`.

src/main/java/org/variantsync/diffdetective/diff/difftree/transform/NaiveMovedArtifactDetection.java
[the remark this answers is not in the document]
pmbittner: I agree.

src/main/java/org/variantsync/diffdetective/diff/difftree/DiffNode.java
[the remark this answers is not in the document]
pmbittner: ok

src/main/java/org/variantsync/diffdetective/diff/difftree/DiffNode.java
```diff
@@ -1112,6 +764,204 @@ public String toTextDiff() {
...
+
+    public static <T> DiffNode unchanged(VariationNode<T> variationNode) {
+        int from = variationNode.getFromLine();
+        int to = variationNode.getToLine();
+
+        var diffNode = new DiffNode(
+            DiffType.NON,
+            variationNode.getNodeType(),
+            new DiffLineNumber(from, from, from),
+            new DiffLineNumber(to, to, to),
```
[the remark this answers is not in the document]
ibbem: This method creates a `DiffNode` from a single `VariationNode` as if diffing the given variation node with itself. Hence, the name `unchanged`.
If the resulting `DiffNode` is printed as diff, there would be no added or removed lines. So the line numbers are all identical.
But I agree that this method needs documentation to clarify that.

src/main/java/org/variantsync/diffdetective/diff/difftree/DiffNode.java
```diff
@@ -1112,6 +764,204 @@ public String toTextDiff() {
...
+     *
+     * <p>Implementation note: It's ensured that identity can be checked using {@code ==}. This
+     * prevents unexpected behaviour if some code uses {@code ==} instead of {@link isSameAs} as
+     * documented in {@link VariationNode}. Although this is currently guaranteed by all
+     * implementations of {@link VariationNode} it should still be considered a bug if {@code ==} is
+     * used to check for identity ({@code null} checks are still allowed).
+     *
+     * @see projection
+     */
+    public class Projection extends VariationNode<DiffNode> {
```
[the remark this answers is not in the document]
ibbem: I agree to extract this class to its own file. I just want to note that the inner class implementation can enforce that only `DiffNode.project` can ever create new `Projection`s (this is important for identity). When extracted the constructor has to be changed to package private so there are more classes with that ability (but which should never instantiate it directly).

src/main/java/org/variantsync/diffdetective/variationtree/VariationNode.java
ibbem: Oh. I think the comment is misleading. I didn't actually write this method so the implementation seems correct (count the number of if-elif-else chains) but I got the comment wrong..

It was previously implemented in `DiffNode` and is currently unused. It was introduced in f72d9552748dcc1fc024c130b6cef391ac016791 (as `getBeforeAnnotationDepth` and `getAfterAnnotationDepth`) and the last usage was removed in 4ac9cdaf3f89675ebd35a1918f52b10fb030a93c (December 2021).

I have no idea if this method will ever be used again. What do you think about removing it?

src/main/java/org/variantsync/diffdetective/variationtree/VariationNode.java
ibbem: This is already checked by `addChild`. Documentation for this behavior is in progress.

src/main/java/org/variantsync/diffdetective/variationtree/VariationNode.java
ibbem: You're right this check is redundant. I just remove it and an exception will be thrown if this invariant isn't met (call of a method on a `null` value) which should be enough for an invariant.

src/main/java/org/variantsync/diffdetective/diff/difftree/DiffNode.java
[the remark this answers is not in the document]
pmbittner: I see. A package-private constructor should be fine though.

src/main/java/org/variantsync/diffdetective/diff/difftree/DiffNode.java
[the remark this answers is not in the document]
pmbittner: Ok I see. Then leave as is for now. One could insert the created DiffNode into a DiffTree though and thus, invalidate the line numbers.

src/main/java/org/variantsync/diffdetective/util/FilterMappedList.java
```diff
@@ -0,0 +1,229 @@
+package org.variantsync.diffdetective.util;
```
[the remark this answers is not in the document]
pmbittner: is merged now

src/main/java/org/variantsync/diffdetective/variationtree/VariationNode.java
pmbittner: I agree to remove it.

src/main/java/org/variantsync/diffdetective/variationtree/VariationTreeNode.java
pmbittner: Ok! Then please leave as is. :)

src/main/java/org/variantsync/diffdetective/diff/difftree/DiffNode.java
```diff
@@ -1112,6 +764,204 @@ public String toTextDiff() {
...
+        private Projection(Time time) {
+            this.time = time;
+        }
+
+        public Time getTime() {
+            return time;
+        }
+
+        @Override
+        public DiffNode getBackingNode() {
```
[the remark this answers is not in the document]
ibbem: Lol. But this is the same term used by the [official java documentation](https://docs.oracle.com/en/java/javase/16/docs/api/java.base/java/util/Collection.html#view).

src/main/java/org/variantsync/diffdetective/diff/difftree/DiffNode.java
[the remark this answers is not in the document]
ibbem: If you mean a different variation diff by *DiffTree*, then yes, but this would be reusing the nodes of one `DiffNode` tree in another (unconnected) `DiffNode` tree. This is always problematic for line numbers, regardless of how they are created.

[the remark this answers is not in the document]
pmbittner: You are right, line numbers get always messed up by tree manipulations. Then keep as is. But I was referring to a DiffTree by DiffTree because a DiffTree is made of DiffNodes, so a DiffNode could be inserted into a DiffTree.

src/main/java/org/variantsync/diffdetective/variationtree/VariationTreeNode.java
ibbem: Thanks for this formulation. I struggled quite a bit to get decent wording and then just gave up.

src/main/java/org/variantsync/diffdetective/variationtree/VariationTreeNode.java
ibbem: Oh. I actually didn't intend this to be a modifiable list as long as we don't actually need it to be.

src/main/java/org/variantsync/diffdetective/variationtree/VariationTreeNode.java
pmbittner: ok :D

src/main/java/org/variantsync/diffdetective/variationtree/VariationTreeNode.java
ibbem: :facepalm: This is the exact same bug as I fixed in #50 (it caused the test failures). This needs to be fixed otherwise `assertConsistency` will detect that and we would have an unexpected inconsistent behavior between `DiffTreen.assertConsistency` and `VariationNode.assertConsistency`.

src/main/java/org/variantsync/diffdetective/diff/difftree/DiffNode.java
```diff
@@ -1112,6 +764,204 @@ public String toTextDiff() {
...
+        Assert.assertTrue(getDiffType().existsAtTime(time));
+
+        if (projections[time.ordinal()] == null) {
+            projections[time.ordinal()] = new Projection(time);
+        }
+
+        return projections[time.ordinal()];
+    }
+
+    public static <T> DiffNode unchanged(VariationNode<T> variationNode) {
```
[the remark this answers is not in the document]
ibbem: Done.

src/main/java/org/variantsync/diffdetective/util/FilterMappedList.java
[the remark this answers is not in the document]
ibbem: I updated Funtjonal and now use their implementation.

src/main/java/org/variantsync/diffdetective/variationtree/VariationNode.java
ibbem: Oh yes. inconsistent with `List` or `DiffNode`? Well I agree being consistent with `DiffNode` is probably better.

src/main/java/org/variantsync/diffdetective/variationtree/VariationTreeNode.java
ibbem: Good catch.

src/main/java/org/variantsync/diffdetective/variationtree/VariationTreeNode.java
ibbem: This would shift the burden of calling `getBackingNode` from the implementation to the caller. That makes the interface considerably less simple.
I currently can't think of new problems that can be cause by mixing such views. I say "new" because all the problems I can imagine are also possible by just mixing two unconnected trees of `DiffNode`s.

After thinking about our discussion about the check in the constructor today, I got another idea:
There is a design pattern in C++ called [curiously recurring template pattern (CRTP)](https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern) where a base class (in our case `VariationNode`) receives the derived class itself (in our case `VariationTreeNode` and `Projection`) as template parameter.
We could use the same design pattern and even enforce it at compile time using the following:
```java
public abstract class VariationNode<T extends VariationNode<T>> implements HasNodeType {
```

This would give us the following benefits:
- We can use the type parameter `T` instead of `VariationNode<T>` everywhere,
- ensuring that no two views are compatible,
- nicer function signatures (though, we should use a different type variable name then) and
- way less invocations of `getBackingNode`.

There are also the following drawbacks:
- Generic functions outside of `VariationNode` have to add a constraint to their type variable:
  before: `public <T> fancyAlgorithm(VariationNode<T> node)`
  after: `public <T extends VariationNode<T>> fancyAlgorithm(T node)`
  This may be confusing for new users. However, I think this is a minor problem as this can be documented and the compiler is able to give a pretty useful error message (giving the correct constraint).

The implementation of this is added as another commit so you can review is separately.

src/main/java/org/variantsync/diffdetective/variationtree/VariationTreeNode.java
ibbem: Well, actually ditching the `getFromLine` and `getAfterLine` methods in favor of only keeping one `getLineRange` method seems pretty nice. It keeps me from writing all such verbose methods and attributes and concentrates it into `Lines`.
For review I implement that proposal and add it as a separate commit. Just tell me if you don't like it.

src/main/java/org/variantsync/diffdetective/variationtree/VariationNode.java
ibbem: See below. I used `Lines` to get rid of all these single line number functions.

src/main/java/org/variantsync/diffdetective/variationtree/VariationNode.java
ibbem: Functions which check their arguments are now documented with `throws IllegalArgumentException`.

src/main/java/org/variantsync/diffdetective/variationtree/VariationNode.java
ibbem: Done.

src/main/java/org/variantsync/diffdetective/variationtree/VariationNode.java
ibbem: See this commit: eb2134152e184d7bd2c245a0f132141e2f2d5eec

src/main/java/org/variantsync/diffdetective/variationtree/VariationTreeNode.java
ibbem: See this commit: eb2134152e184d7bd2c245a0f132141e2f2d5eec

src/main/java/org/variantsync/diffdetective/variationtree/VariationTreeNode.java
ibbem: See this commit: eb2134152e184d7bd2c245a0f132141e2f2d5eec

src/main/java/org/variantsync/diffdetective/variationtree/VariationTreeNode.java
ibbem: See this commit: 1cb04cecebcbcb7426d576f8716c0dafa6e31ef9

src/main/java/org/variantsync/diffdetective/variationtree/VariationNode.java
pmbittner: Yes I would also prefer being consistent with DiffNode.

src/main/java/org/variantsync/diffdetective/variationtree/VariationNode.java
pmbittner: Good idea!

src/main/java/org/variantsync/diffdetective/variationtree/VariationTreeNode.java
pmbittner: Good idea! I added some comments below.

src/main/java/org/variantsync/diffdetective/variationtree/VariationTreeNode.java
pmbittner: I actually think the code is now easier to read because logic for data and view is not mixed anymore. In particular, we do not have `Projection extends VariationNode<DiffNode>` anymore which involves three types.

src/main/java/org/variantsync/diffdetective/variationtree/VariationTreeNode.java
pmbittner: I like it. We only have to be careful with the variables now being references to objects and no primitive values anymore. I had some bugs with that in the past but it seems to be as if you have been careful here, by making the set-methods side-effect free in terms of returning a new value.

pmbittner (changes requested):

src/main/java/org/variantsync/diffdetective/variationtree/VariationNode.java
```diff
@@ -0,0 +1,545 @@
...
+        }
+
+        return List.of(getDirectFeatureMapping());
+    }
+
+    /**
+     * Returns the full feature mapping formula of this node.
+     *
+     * <p>The feature mapping of an {@link NodeType#IF} node is its
+     * {@link DiffNode#getDirectFeatureMapping direct feature mapping}. The feature mapping of
```
pmbittner, line 336: wrong reference to DiffNode. Should be VariationNode right?

src/main/java/org/variantsync/diffdetective/variationtree/VariationNode.java
```diff
@@ -0,0 +1,545 @@
...
+     * {@link NodeType#ELSE} and {@link NodeType#ELIF} nodes is determined by all formulas in the
+     * respective if-elif-else chain. The feature mapping of an {@link NodeType#ARTIFACT artifact}
+     * node is the feature mapping of its parent.
+     * See Equation (1) in <a href="https://doi.org/10.1145/3540250.3549108">our paper</a>.
```
pmbittner, line 340: Use link to preprint for open access https://github.com/SoftVarE-Group/Papers/blob/raw/2022/2022-ESECFSE-Bittner.pdf

src/main/java/org/variantsync/diffdetective/variationtree/VariationTreeNode.java
pmbittner: Maybe we should document in the method's javadoc which information is lost and only recovered by heuristics (i.e., line numbers and formula of annotations)?

src/main/java/org/variantsync/diffdetective/variationtree/VariationTreeNode.java
ibbem: Both `Lines` and `DiffLineNumber` are immutable to this should no longer be a problem.

src/main/java/org/variantsync/diffdetective/variationtree/VariationNode.java
ibbem: yes

src/main/java/org/variantsync/diffdetective/variationtree/VariationTreeNode.java
```diff
@@ -137,23 +130,21 @@ public VariationTreeNode(
     public static VariationTreeNode createRoot() {
         return new VariationTreeNode(
                 NodeType.IF,
-                DiffLineNumber.InvalidLineNumber,
-                DiffLineNumber.InvalidLineNumber,
                 FixTrueFalse.True,
+                null,
```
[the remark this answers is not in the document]
ibbem: You're right `Lines.Invalid()` is probably better here.

src/main/java/org/variantsync/diffdetective/variationtree/VariationTreeNode.java
ibbem: I don't think it's a good idea to give definite answers to what exactly can be recovered from this ID because this would encourage using it for serialization. Misusing `getID` for serializing some fields of `DiffNode` was a mistake in my opinion. It's only real use case should be identifying a node uniquely in a tree.
It's probably a good idea to completely eliminate `fromID` in the future as it's not really stable across versions (I already changed the IDs multiple times) and implicitly looses information.

pmbittner: I see your point. So the misuse would be any use of `fromID` but `toID` should be fine. I implemented these methods because of a use case were we had to export DiffTrees for another tool that reads linegraph files which in turn require natural number identifiers for nodes. I agree that upon serialisation we should either use a proper format that stores node data separately (e.g., in a json or in a text label in the linegraph file).

src/main/java/org/variantsync/diffdetective/variationtree/VariationTreeNode.java
pmbittner: Yes! :)

pmbittner (approved):

pmbittner:
Thank you for all the fixes and adaptations. We can merge now.  🥳

PR #63, opened 2023-01-17
Integration of LB Thesis

pmbittner:


ibbem:
There are some issues which I fixed:
- It doesn't actually compile
  - b5b0b00fe3effe9dad9dda01226cea2c3336a82e Remove unused code which has missing imports
  - 74b143d4f89fae98768bf1ec85122c05f06e93b6 Remove duplicated code causing compilation failures

- Filesystem cleanup
  - 4462c703b2f38a97198f854d3096842b881d4d33 Remove and ignore VS Code specific files
  - bb3ce4bdab9aa35c4d270b489da93e302f80e576 Move all datasets into the `docs/datasets` directory

- Formatting cleanup
  - 53e0948e7d220fc2f9591b35055965a6caa792e0 Remove already fixed TODOs
  - baa0e61f7be17d81413d9d66938b1ed20921697b Remove unused imports
  - 17fc822e253a3d51bf8295ed06427991cb9124b4 Fix whitespace issues

- Simplifications/Refactorings
  - 7250bd5a07a63919c717c0007937e5670a0a63d8 Refactor `DiffEdge` to a record
  - f7815cbddeb09ed30b51b7d308efa864dc37aa01 Refactor the duplication code for `DiffNode`s
  - 43e89c7cf437dc4dd2bdaf1e5cf6441cde3860e1 Move the duplication code to `DiffTree` and `DiffNode`
  - a1bc144c6383de92f61903f5c837fc828df92314 Refactor AtomicDiffComparison
  - 8ae76865c1df2336a20800aa5359dbf60393dd35 Simplify an ArrayList creation
  - 144570e51f93a92fa6dbe2f7e3caa6f8e30a904e Refactor `FeatureSplit`

- Removal of duplicate code
  - 88fbcedf1801708b6f3c86af7903ef47ebd571a0 Factor duplicate code out of `*Analysis` and `*Result`
  - 70c6e290f6b37f350e40de6e4c7bd841c05b7141 Merge `FeatureSplitAnalysisTask` and `CommitHistoryAnalysisTask`
  - 111b7150e5ab808e3de0033c90cdb2c432b32ef7 Rename `Validation` to `EditClassValidation`
  - af6afe5eb502e25f62a40abbe506396d0654c747 Factor duplicate code out of the `validation` package
  - ee6964a982714dd920a7a46fc117615a46dea9b3 Remove duplicated test

- Misc (Not all related to this PR but I stumbled upon these while reviewing)
  - 019ffe7c07636e0df29925687b8d2aa71ad19e36 Escape all shell variable substitutions
  - ab4acb4a78fab5bcb7a192ba6983fe1e5d9b4854 Document a precondition in `CPPDiffLineFormulaExtractor`
  - 936015bd926b1e629e85746b711bdcc9fffc6bbd Imrove log messages in `Analysis.forEachRepository`

There is some more potential to remove duplicate code from the `*Task` classes. I want to handle these as soon as the changes from #53 are integrated with this PR. This hopefully allows for a better interface for the simplifications.

ibbem:
As requested, I merged develop into this branch.
I continued to work on the analysis/validation refactoring, so sorry for the big diff :).
Of course, there is still cleanup potential but now I am quite happy about the general structure of the analysis/validation.

ibbem:
@pmbittner ping for review

src/main/java/org/variantsync/diffdetective/analysis/LineGraphExportAnalysis.java
```diff
@@ -0,0 +1,91 @@
...
+public class LineGraphExportAnalysis implements Analysis.Hooks {
+    public static final ResultKey<Result> RESULT = new ResultKey<>("LineGraphExportAnalysis");
+    public static final class Result implements Metadata<Result> {
+        public String treeFormat;
+        public String nodeFormat;
+        public String edgeFormat;
+
+        public static final InplaceSemigroup<Result> ISEMIGROUP = (a, b) -> {
+
+        };
```
pmbittner, line 22: Throw an exception if the formats of a and b are different? Only equal formats should be merged right?

src/main/java/org/variantsync/diffdetective/analysis/Analysis.java
```diff
@@ -0,0 +1,351 @@
...
+        while (hook.hasNext()) {
+            callHook.accept(hook.next(), this);
+        }
+    }
+
+    protected <Hook> boolean runFilterHook(ListIterator<Hook> hook, FailableBiFunction<Hook, Analysis, Boolean, Exception> callHook) throws Exception {
+        while (hook.hasNext()) {
+            if (!callHook.apply(hook.next(), this)) {
+                return false;
+            }
```
pmbittner, line 308: FilterHooks should be side-effect free but they aren't. For example, `LineGraphExportAnalysis::onParsedCommit` has a side-effect. When exitiving here as soon as some filter says "no", we would omit running the remaining side-effects = bug! So I would argue to just run all hooks and make a big AND of their return values.

src/main/java/org/variantsync/diffdetective/analysis/PatchAnalysis.java
```diff
@@ -0,0 +1,74 @@
...
+
+import org.tinylog.Logger;
+import org.variantsync.diffdetective.editclass.EditClass;
+import org.variantsync.diffdetective.editclass.proposed.ProposedEditClasses;
+import org.variantsync.diffdetective.metadata.EditClassCount;
+import org.variantsync.diffdetective.util.CSV;
+import org.variantsync.diffdetective.util.FileUtils;
+import org.variantsync.diffdetective.util.IO;
+
+public class PatchAnalysis implements Analysis.Hooks {
```
pmbittner, line 16: I see why you chose this name but it is misleading. This analyses counts only edit class occurences so name it `EditClassOccurenceAnalysis`?

pmbittner: This would probably also mean to rename the `PatchStatistics` to something like `PatchEditClassRecord`?

src/main/java/org/variantsync/diffdetective/analysis/PreprocessingAnalysis.java
```diff
@@ -0,0 +1,25 @@
...
+        this.preprocessors = Arrays.asList(preprocessors);
+    }
+
+    @Override
+    public boolean analyzeDiffTree(Analysis analysis) throws Exception {
+        DiffTreeTransformer.apply(preprocessors, analysis.getDiffTree());
+        analysis.getDiffTree().assertConsistency();
+        return true;
+    }
+}
```
pmbittner, line 25: nice!

src/main/java/org/variantsync/diffdetective/feature/CPPDiffLineFormulaExtractor.java
```diff
@@ -44,6 +44,15 @@ public String extractFormula(final String line) throws IllFormedAnnotationExcept
         final Supplier<IllFormedAnnotationException> couldNotExtractFormula = () ->
                 IllFormedAnnotationException.IfWithoutCondition("Could not extract formula from line \""+ line + "\".");
 
+        // check for equal number of brackets
+        // The C11 standard actually allows character literals so there could be `')'` in a
+        // condition but that's never used in practice, right?
```
pmbittner, line 49: Indeed it would be weird having these in CPP conditions.

src/main/java/org/variantsync/diffdetective/feature/CPPDiffLineFormulaExtractor.java
```diff
@@ -65,6 +74,13 @@ public String extractFormula(final String line) throws IllFormedAnnotationExcept
         // remove whitespace
         fm = fm.replaceAll("\\s", "");
 
+        if (fm.isEmpty()) {
+            throw couldNotExtractFormula.get();
+        }
+
+        // remove defined()
+        fm = DEFINED_PATTERN.matcher(fm).replaceAll("$1");
+        fm = fm.replaceAll("defined ", " ");
```
pmbittner, line 83: Are these changes new? Did Lukas Bormann make these or you @ibbem ? Also `defined(x)` has different semantics in CPP than `x` so maybe we should not drop `defined` clauses but rather make `x` become `DEFINED_x`?

src/main/java/org/variantsync/diffdetective/tablegen/MiningResultAccumulator.java
```diff
@@ -21,26 +23,16 @@
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.function.BiConsumer;
 import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
 /** Accumulates multiple {@link AnalysisResult}s of several datasets. */
 public class MiningResultAccumulator {
```
pmbittner, line 31: Rename to `AnalysesResultAccumulator`?

src/main/java/org/variantsync/diffdetective/validation/EditClassValidation.java
```diff
@@ -0,0 +1,70 @@
...
+        "EditClassValidation",
+        List.of(
+            new PreprocessingAnalysis(new CutNonEditedSubtrees()),
+            new FilterAnalysis(DiffTreeFilter.notEmpty()), // filters unwanted trees
+            new EditClassValidation(),
+            new StatisticsAnalysis()
+        ),
+        repo,
+        repoOutputDir
+    );
```
pmbittner, line 36: This new compositional approach is indeed quite nice! Good job!

pmbittner (commented):
Hi @ibbem ,

thanks for the heavy refactoring of the analyses framework. It looks much better now. It was not easy to review it given the extensive amount of changes and the still complicated analyses we perform.

Two major things I noticed are:
1.) I did not review the FeatureSplit parts yet. I have to dig into those in the next weeks. Can you move all the new FeatureSplit related classes to dedicated sub-packages so that we can easily identify them after the merge? What is your opinion on the location of the new code?
2.) Please add some documentation (javadoc) to the new analyses framework. It is still complicated and guidance for developers on how to properly use it is missing.

After that we can merge. Thank you. :)

src/main/java/org/variantsync/diffdetective/analysis/Analysis.java
```diff
@@ -0,0 +1,351 @@
...
+    protected void processCommitBatch(List<RevCommit> commits) throws Exception {
+        outputFile = outputDir.resolve(commits.get(0).getId().getName() + ".lg");
+
+        ListIterator<Hooks> batchHook = hooks.listIterator();
+        try {
+            runHook(batchHook, Hooks::beginBatch);
+
+            // For each commit
+            for (final RevCommit finalCommit : commits) {
+                commit = finalCommit;
```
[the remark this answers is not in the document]
ibbem: I'm also not very happy about it but each hook needs successively more state (there are currently eight state variables). The only reason that `Analysis` can be instantiated is to pass this state around more easily.
You're correct about the thread safety. Each batch (thread) instantiates a new `Analysis` without sharing any fields except the `GitDiffer` (it was already shared before this refactoring and I preserved that behaviour).
All publicly accessible fields in `Analysis` have the semantic of only being valid during the current hook. Should I prefix all fields and their getters with `current`?

src/main/java/org/variantsync/diffdetective/analysis/Analysis.java
```diff
@@ -0,0 +1,351 @@
...
+            } catch (Exception e) {
+                Logger.error(e, "An exception thrown in an end hooks of Analysis will be rethrown later");
+                catchedException = e;
+            }
+        }
+
+        if (catchedException != null) {
+            throw catchedException;
+        }
+    }
```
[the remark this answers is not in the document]
ibbem: Some hooks need to throw checked exceptions (e.g. `IOException`s). There is no reason to expect that any hook can't throw certain kinds of exceptions, therefore all methods in `Analysis.Hooks` have to throw generic `Exception`s. Thus I didn't bother to add type parameters to `run*Hook` as there is currently no benefit in doing that.

src/main/java/org/variantsync/diffdetective/analysis/Analysis.java
```diff
@@ -0,0 +1,351 @@
...
+
+    protected <Hook> void runReverseHook(ListIterator<Hook> hook, FailableBiConsumer<Hook, Analysis, Exception> callHook) throws Exception {
+        Exception catchedException = null;
+        while (hook.hasPrevious()) {
+            try {
+                callHook.accept(hook.previous(), this);
+            } catch (Exception e) {
+                Logger.error(e, "An exception thrown in an end hooks of Analysis will be rethrown later");
+                catchedException = e;
```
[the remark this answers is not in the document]
ibbem: Hmm. I didn't know about `CompositeException`. How am I supposed to instantiate it? Is really intended to be used like this:
```java
CompositeException catchedException = null;
while (hook.hasPrevious()) {
    try {
        callHook.accept(hook.previous(), this);
    } catch (Exception e) {
        Logger.error(e, "An exception thrown in an end hooks of Analysis will be rethrown later");
        if (catchedException == null) {
            catchedException = new CompositeException(e);
        } else {
            catchedException = new CompositeException(catchedException, new CompositeException(e));
        }
    }
}

if (catchedException != null) {
    throw catchedException;
}
```

I just checked the official Java documentation: There is a method [`Throwable.addSuppressed`](https://docs.oracle.com/en/java/javase/16/docs/api/java.base/java/lang/Throwable.html#addSuppressed(java.lang.Throwable)) which seems to do exactly what we need in this case:
```java
Exception catchedException = null;
while (hook.hasPrevious()) {
    try {
        callHook.accept(hook.previous(), this);
    } catch (Exception e) {
        Logger.error(e, "An exception thrown in an end hooks of Analysis will be rethrown later");
        if (catchedException == null) {
            catchedException = e;
        } else {
            catchedException.addSuppressed(e);
        }
    }
}

if (catchedException != null) {
    throw catchedException;
}

```

What do you think about `Throwable.addSuppressed`?

src/main/java/org/variantsync/diffdetective/analysis/LineGraphExportAnalysis.java
ibbem: Yes, something like
```java
if (b.treeFormat != null) {
    if (a.treeFormat != null) {
        Assert.assertTrue(a.treeFormat.equals(b.treeFormat));
    } else {
        a.treeFormat = b.treeFormat;
    }
}
```
would work. In this case we should check every semigroup instance for similar cases. Of the top of my head, I know at least one other case which requires different behavior: `AnalysisResult.repoName` can't throw an exception in such a case as it will be different in `Analysis.forEachCommit`.

src/main/java/org/variantsync/diffdetective/analysis/Analysis.java
ibbem: No. Filter hooks are not required to be side-effect free. I will explain this in the upcoming documentation.

src/main/java/org/variantsync/diffdetective/analysis/PatchAnalysis.java
ibbem: Sounds reasonable.

Because of that new name, I just spotted duplicate code: `EditClassValidation.analyzeDiffTree` and `PatchAnalysis.analyzeDiffTree` do the same thing except `PatchAnalysis.analyzeDiffTree` also stored the edit class occurrences in patch granularity. Should we do something abut this or is this acceptable?

src/main/java/org/variantsync/diffdetective/validation/EditClassValidation.java
ibbem: thanks

src/main/java/org/variantsync/diffdetective/feature/CPPDiffLineFormulaExtractor.java
ibbem: Yes, these changes are new but there where already patches like this in #50. These are essentially duplicates which I missed during merging and should be removed. Indeed, the defined operators are just ignored since at least 9b5e59c2a2ea9a67be34e64311482e562237a401.
Adding a `DEFINED_` prefix seems good to me. I will add it soon.

src/main/java/org/variantsync/diffdetective/tablegen/MiningResultAccumulator.java
ibbem: I don't think this is a better name for the current implementation as it doesn't accumulate all results but only `ExplainedFilterSummary`, `EditClassCount` and `StatisticsAnalysis` which are created by the `DiffTreeMiner`.
We could generalize `MiningResultAccumulator` by either
1. creating a registry for all `Metadata` so it can be created in `AnalysesResultAccumulator`, or
2. adding the qualified class names to the serialization and creating the relevant classes at runtime.

Then `AnalysesResultAccumulator` would be appropriate.

ibbem:
1. I don't think it's a good idea to move all feature split related operations into a sub package. To me, it does make more sense to have a package with all transformation we can perform on `DiffTree`s as we have now.
2. I am currently writing documentation. Stay tuned :)

During discussion you noted that `thesis_lb` should be essentially freezed and be reset to before the `origin/develop` merge and all these refactorings. As we loose all the above discussion I will postpone this action until the discussion is finished.

ibbem:
I integrated the changes into `benjamin/analysis` (#65) and `thesis_lb-refactorings`. For comparison I uploaded the state before these changes but after the rebase onto develop to `benjamin/old/analysis-refactoring` (Note: I made some cosmetic changes during the rebase to make the single commits more readable).

src/main/java/org/variantsync/diffdetective/analysis/Analysis.java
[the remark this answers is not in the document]
ibbem: You can see the `Throwable.addSuppressed` approach in #65.

src/main/java/org/variantsync/diffdetective/analysis/LineGraphExportAnalysis.java
ibbem: Integrated into #65, for all fields noted above.

src/main/java/org/variantsync/diffdetective/analysis/PatchAnalysis.java
ibbem: As the name `EditClassCounts` was overloaded (in `metadata` and `analysis`) I merged `EditClassCounts` into `PatchStatistics` and further simplified `EditClassOccurenceAnalysis` by inlining the result. It now also uses `AnalysisStrategy` for the export.

src/main/java/org/variantsync/diffdetective/analysis/Analysis.java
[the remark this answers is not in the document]
pmbittner: ok

src/main/java/org/variantsync/diffdetective/analysis/Analysis.java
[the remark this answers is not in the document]
pmbittner: Ok. Please document said semantics. I think prefixing `current` would help emphasize these semantics apart from pure documentation.

src/main/java/org/variantsync/diffdetective/analysis/Analysis.java
pmbittner: As discussed in our last meeting, the current behaviour is fine but should be documented.

src/main/java/org/variantsync/diffdetective/analysis/PatchAnalysis.java
pmbittner: I don't remember all the details here. I wouldn't be surprised If introduced quite some duplicate code here. Refactoring is fine and appreciated as long as it indeed is a refactoring.

src/main/java/org/variantsync/diffdetective/feature/CPPDiffLineFormulaExtractor.java
pmbittner: Ok good! So you removed this change for #65 ? You plan on opening a new pull-request for the DEFINED-issue in the future? Maybe opening an issue as a reminder would be good but I leave that up to you.

src/main/java/org/variantsync/diffdetective/tablegen/MiningResultAccumulator.java
pmbittner: Let's discuss that in a meeting. Putting effort in here only makes sense if we actually profit from such a refactoring.

src/main/java/org/variantsync/diffdetective/analysis/Analysis.java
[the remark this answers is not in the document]
ibbem: OK, I'll use `Throwable.addSuppressed` then.

Regarding Funtjonal:
The problem with `CompositeException` is that there is no access to the inner list and the only constructor accepting more than one exception only accepts `CompositeException`s as a variable number of arguments. There is no way a `List<Exception>` can be passed to `CompositeException` as would be needed here. I could open an issue in Functjonal if this looks like desired behavior to you.

src/main/java/org/variantsync/diffdetective/feature/CPPDiffLineFormulaExtractor.java
ibbem: Yes, this change is not included in #65. The change was already finished. See #66.

src/main/java/org/variantsync/diffdetective/analysis/Analysis.java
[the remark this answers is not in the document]
ibbem: OK, I prefixed all relevant state variables an getters with `current`.

src/main/java/org/variantsync/diffdetective/analysis/Analysis.java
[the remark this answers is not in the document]
pmbittner: It is fine, don't bother with it now.

pmbittner:
#65 has been merged. Closing this PR now.

PR #64, opened 2023-01-18
Reorganize the package structure related to variation trees/diffs

ibbem:


pmbittner:
Changes look good! Maybe we should somehow indicate that the `diff` package is about variability-unaware text-based diffs and git wrappers. We could add some subpackages for that (e.g., `diff.text` and `diff.text`)? What do you think?

pmbittner (approved):

ibbem:
Yes, that works out quite nice. Although the separation between `diff.git` and `diff.text` is currently not as clear cut (e.g. `DiffFilter` should be in `diff.text` but currently depends on Git) but that can be dealt with when necessary.
So this can be merged, right? @pmbittner

PR #65, opened 2023-02-20
Analysis refactoring

ibbem:
Refactoring of the analysis machinery. These changes where motivated by the changes in the `thesis_lb` branch and where originally discussed in #63. Note that there are some unrelated changes because of this.


ibbem:
Note that 8370477ab42aed88f8e615432f3f09bacf05ae84 is an artifact of #63 where it was initially introduced by TheBormann.

src/main/java/org/variantsync/diffdetective/metadata/Metadata.java
```diff
@@ -37,6 +40,18 @@ default void append(T other) {
         semigroup().appendToFirst(Cast.unchecked(this), other);
     }
 
+    static <T> T mergeEqual(T a, T b) {
```
pmbittner, line 43: Document behaviour and intention?

pmbittner (approved):
Only one comment and afterwards we can merge.

ibbem:
Added documentation for `Metadata.mergeEqual`, `AnalysisResult.ResultKey`, `AnalysisResult.get` and `AnalysisResult.append`.

PR #66, opened 2023-02-21
Differentiate between #if A and #if defined(A)

ibbem:
This patch relaxes the CPP formula parser to also accept `#if defined()` although it has an invalid formula. It would make the code very ugly to test for this case so I left it out.


pmbittner:
Changes look good!

You mean we would parse `#if defined(A && B)` to `DEFINED_A_AND_B` or something like that? What about refining the regex for finding `defined` to allow for only a single named argument? `defined` + whitespace* + `(` + whitespace* + identifier + whitespace* + `)`?

ibbem:
`#if defined(A && B)` is invalid CPP syntax. It would be transformed to `DEFINED_A&&B` (the `&&` is parsed by FeatureIDE). 

The current regex is almost exactly what you described: `\bdefined\b(\s*\((\w*)\))?`.
Each version has a bug :).
It should actually be `\bdefined\b(\s*\((\s*\w*\s*)\))?`

PR #72, opened 2023-03-07
Online rendering of Variation Trees and Diffs for Live Previews

pmbittner:
This pull-request adds a package for rendering variation trees and diffs to a window that can be opened at any point during execution. The workflow is inspired by python's matplotlib: At any point during execution, a call to
```java
//...
Show.diff(someVariationDiff).showAndAwait();
//...
```
will halt the execution and open a window displaying the given diff or tree:
grafik
When the window is closed, execution will continue. Alternatively, the window can also be opened to run in parallel with the main application, thus not causing the control flow to halt.

This feature is intended for debugging and demonstrations. So far it is solely a viewer and _not_ and editor for graphs.

The windows features are:
- zoom (mouse wheel)
- scroll across the view (hold and drag right mouse button)
- drag and drop of nodes to adjust layout (hold and drag nodes with left mouse button)
- switch the layout of nodes (via `Layout` menu)
- switch the labels of nodes (via `Labels` menu)
- save as screenshot (CTRL+S or via `File` menu)
- show root, which is useful for large diffs (R key)

The code changes comprise:
- A lightweight game engine to (1) describe a game world containing entities, with (2) useful linear algebra functions, and (3) facilities to render the world state to (4) a window.
- One domain-specific application for said game engine that displays variation trees or diffs.
- Option to render to texture without ever opening a window. This could replace the old python based rendering.
- Functionality to display multiple windows at once in a certain layout. This is useful when debugging code that manipulates graphs.
- A refactoring of `DiffNode::unchanged` to allow custom conversion from `VariationTreeNode`s to `DiffNode`s. This is helpful for converting ´BadVDiffs´ to `DiffTree`s.
- A refactoring to the node layouting procedure with graphviz used by the `TikzExporter` and provided by the `GraphvizExporter`. Now it is possible to obtain node layouts irrespective of a tikz export. We use this to layout the nodes in the show view.
- Renaming `LinegraphFormat::getName()` to `LinegraphFormat::getIdentifier()` since that method actually yields the fully qualified class name of the format, which turned out to be cumbersome for display in the show view. Instead, `LinegraphFormat` now has a second method `getShortName()` that returns only the class name.
- Two new `DiffNodeLabelFormat`s that are useful for the show view.
- `DiffTree::getNodeWithID`
- `Assert::assertNull`


pmbittner:
Currently, the `ExportTest::export()` fails because Graphiviz produces slightly different coordinates on my machine:
grafik
This test seems to be fragile w.r.t. the installed graphviz version and its configuration.

PR #73, opened 2023-03-07
Some old updates and refactorings to the tikz exporter

pmbittner:
Hi @ibbem,

these are some tweaks I made to the tikz exporter a few months ago, when I used it to generate graphics for my presentation.

If I remember correctly, the major change here was to treat parallel edges as one unchanged edge during the export. Otherwise, the two parallel edges were drawn on top of each other and only one of them remained visible.

This PR also adds a new main method to have a small program that turns text-diffs to tikz. It might be a good idea to extend that main method later with the `Show` facilities from PR #72 . Then, we could have a small utility program that loads a given text diff in our view which lets you choose a proper layout and manually rearrange nodes, and then export to tikz.

### DIFF
--- a/src/main/java/org/variantsync/diffdetective/internal/TextDiffToTikz.java
+++ b/src/main/java/org/variantsync/diffdetective/internal/TextDiffToTikz.java
@@ -1,0 +1,109 @@
+package org.variantsync.diffdetective.internal;
+
+import org.tinylog.Logger;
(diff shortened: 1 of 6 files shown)


pmbittner:
@ibbem If you agree with these changes, could you have a look at the failing unit tests? The changes to the tikz exporter invalidated the unit tests. :(

PR #75, opened 2023-03-09
Bugfix Regarding Empty Lines in Diffs

pmbittner:
- Treat empty lines as unchanged lines.
- The parser did not increase the line number counter when it skipped an empty line. This resulted in wrong line numbers for all subsequently parsed lines.


ibbem:
The commit 8ec9b535 looks good.

I think the commit 43836067 should not be applied as diffs without a character in the first column of a line is malformed (unchanged lines must have a space).

I close this pull request in favor of #78 to prevent the commit history from getting even more complicated because all of this was already (indirectly) merged into #77 and #78 . The commit 43836067 will have to be reverted using `git revert 43836067`.

PR #76, opened 2023-03-10
Make collapseMultipleCodeLines and ignoreEmptyLines configurable

pmbittner:
These two parameters were previously hardcoded to `true` within the `GitDiffer` when parsing all `DiffTree`s within a commit.


ibbem:
The commits 9877979b, e9582b9a and 56c8b681 look good to me.

I found a two minor issues in fbc05e46:
- Missing `@param` in `PatchDiffParseOptions`. I think we should either document all parameters or none (if their meaning is obvious).
- I don't think `ignoreEmptyLines = true` and `collapseMultipleCodeLines = true` are good defaults for `DiffTreeParseOptions`. At some point we actually discussed eliminating `ignoreEmptyLines` thus `ignoreEmptyLines = false` wouldn't break code using the defaults. Similarly, `collapseMultipleCodeLines` could be extracted to a `DiffTreeTransformer` to untangle the parser a little bit thus the same reasoning applies.

I close this pull request in favor of #78 to prevent the commit history from getting even more complicated because all of this was already (indirectly) merged into https://github.com/VariantSync/DiffDetective/pull/77 and https://github.com/VariantSync/DiffDetective/pull/78.

PR #77, opened 2023-03-10
Enable Parsing of Inserted and Deleted Files

pmbittner:
Previously, parsing patches that inserted a file (not modify) failed because no file before the patch could be found to embed the patch in. However, embedding the patch in a before file is unnecessary (and impossible) when the file was inserted. In this case, the patch itself is already the full diff to parse the variation diff from. This PR fixes that.


pmbittner:
Merge after ##76 and #75

ibbem:
There is one minor issue in the commit 5f13ec2e:
In line  334 of the `GitDiffer` there is inefficient string processing:
```java
strippedDiff.lines().skip(1).collect(Collectors.joining(StringUtils.LINEBREAK));
```
This splits the whole file into lines and rejoins all but the first of them. This operation is quite expensive for big files considering how little it actually does. We should refactor this at some point.

I close this pull request in favor of #78 to prevent the commit history from getting even more complicated because all of this was already merged into #78.

PR #78, opened 2023-03-10
Parsing of Commits Without Parents (e.g., Initial Commit)

pmbittner:
Merge after #77


ibbem:
The commit c8163639 looks good to me.

The commit 8506be72 contains some minor issues:
- There is an unnessecary nested catch in line 230 of the `GitDiffer`.
- We could use the more specific `CanonicalTreeParser` constructor instead of `reset` to simplify the code around line 233 and 285 in the `GitDiffer`

### Recap of #75, #76 and #77
The commits 8ec9b535, 9877979b, e9582b9a and 56c8b681 look good to me.

I think the commit 43836067 should not be applied as diffs without a character in the first column of a line is malformed (unchanged lines must have a space).

I found a two minor issues in fbc05e46:
- Missing `@param` in `PatchDiffParseOptions`. I think we should either document all parameters or none (if their meaning is obvious).
- I don't think `ignoreEmptyLines = true` and `collapseMultipleCodeLines = true` are good defaults for `DiffTreeParseOptions`. At some point we actually discussed eliminating `ignoreEmptyLines` thus `ignoreEmptyLines = false` wouldn't break code using the defaults. Similarly, `collapseMultipleCodeLines` could be extracted to a `DiffTreeTransformer` to untangle the parser a little bit thus the same reasoning applies.

There is one minor issue in the commit 5f13ec2e:
In line  334 of the `GitDiffer` there is inefficient string processing:
```java
strippedDiff.lines().skip(1).collect(Collectors.joining(StringUtils.LINEBREAK));
```
This splits the whole file into lines and rejoins all but the first of them. This operation is quite expensive for big files considering how little it actually does. We should refactor this at some point.

### Next actions
The commit 43836067 should be reverted using `git revert 43836067`. Then this pull request can be merged (this merges all the code of #75, #76 and #77 as these are already merge into this PR). I think we could deal with all of these minor issues in further refactorings (Except maybe the change of defaults).

ibbem (approved):

pmbittner:
Thank you for the detailed review and taking care of the history. I agree with all your comments. For the defaults of in fbc05e46, I am ok with changing the defaults as proposed. For the FSE'22 EditClassValidation we should stick to the values used back then though; but this is not a problem because now that these options are configurable, the EditClassValidation can just set those values explicitly.

pmbittner:
I guess its easier to apply these changes now rather than later and having to open issues for them.

pmbittner:
Could you apply the fixes?

pmbittner:
fixed :)

pmbittner:
Rebased onto `develop`.

src/main/java/org/variantsync/diffdetective/diff/git/GitDiffer.java
```diff
@@ -317,18 +327,48 @@ private static CommitDiffResult getPatchDiffs(
...
+                }
+
                 try {
-                    final BufferedReader file = getBeforeFullFile(git, parentCommit, filename);
-                    final PatchDiff patchDiff =
-                        createPatchDiff(
+                    final String fullDiff = switch (diffEntry.getChangeType()) {
+                        case ADD, DELETE -> {
+                            // The first lines contains meta information "@@ ... " that we want to skip.
+                            final String[] hunkBeginAndRest = strippedDiff.split(StringUtils.LINEBREAK_REGEX_RAW, 2);
```
ibbem, line 343: You can use `StringUtils.LINEBREAK_REGEX.split(strippedDiff, 2)` to avoid introducing `LINEBREAK_REGEX_RAW` and recompiling the regex again.

ibbem (changes requested):
Thanks for cleaning up the history with a rebase. Now everything looks nice (except one minor comment).

PR #80, opened 2023-03-17
Store exact line numbers for all lines

ibbem:
The exact line numbers are now stored in the label of `DiffNode`s. Furthermore, `#endif`s are stored in the label of the preceding preprocessor directive belonging to the same if-chain. Note that there can be two different `#endif` lines in the same label if the `#endif` changed. In this case the line number of each will only be valid either before or after the edit respectively. See `src/test/resources/diffs/parser/09.diff` and `src/test/resources/diffs/parser/16.diff` for examples.

It should now be possible to extract the full file before and after the edit by iterating over the variation diff and collecting all label lines existing at the required time and ordering by their line number. Unfortunately, we don't have tests for this but we plan to implement an unparse method for creating a full diff from a variation diff.

Alternatively to storing `#endif`s in the labels, we could reintroduce an endif `NodeType` and create extra nodes for `#endif`s. Furthermore, it may be necessary to make the addition of `#endif`s configurable.

The labels and especially the label interactions between variation trees and variation diffs got quite complex (and variation trees do not even store line numbers currently). Thus, it may be necessary to use generics to allow different types of labels in `DiffNode`s and `VariationNode`s. Such a change would be quite huge and break most user code, so I propose to combine this change with the planned rename of `DiffTree` to something like `VariationDiff` as that would have similar consequences.


ibbem:
As CI doesn't run on this PR everything seems good, but there are still some invalid Javadoc references. These are fixed by 56c8b681adcfffc29cccb3810c471422fd9969a3 so I decided to ignore them to reduce merge conflicts.

ibbem:
I'm a bit unsure onto which branch I should base these changes. Is the branch `as` supposed to be a feature branch or a branch owned by AlexanderSchultheiss for deviations required for his use case? In other words: Should we this PR into `as` and later `as` into `develop` or should we merge this PR into `develop` and regularly merge `develop` into `as`?

pmbittner:
I appreciate the idea of making the labels of nodes more generic and thus exchangeable. This would make nodes become more close to their mathematical definition of just holding a value of some type. I agree that this would be a big breaking change though.

The idea to group it with the renaming of `DiffTree` to `VariationDiff` sound great to me. I would propose the following:
1.) Merge the current PRs into develop.
2.) Ensure tests to pass and test the EditClassValidation.
3.) Merge in main and name it version 1.0.
4.) Apply breaking changes in new PRs to develop. When we merge the breaking changes from develop later to main, we can increase the big version number to 2.0.

What do you think @ibbem ?

src/main/java/org/variantsync/diffdetective/variation/tree/VariationNode.java
```diff
@@ -437,7 +450,7 @@ public VariationTreeNode toVariationTree() {
             getNodeType(),
             getFormula().clone(),
             getLineRange(),
-            new ArrayList<String>(getLabelLines())
+            getLabel().lines()
```
pmbittner, line 453: Why remove the copying? We still clone te formula.

src/main/java/org/variantsync/diffdetective/variation/diff/parse/DiffTreeParser.java
```diff
@@ -315,7 +315,7 @@ private void parseLine(
 
         // Is this line a conditional macro?
         // Note: The following line doesn't handle comments and line continuations correctly.
-        var matcher = macroPattern.matcher(line.getLines().get(0));
+        var matcher = macroPattern.matcher(line.toString());
```
pmbittner, line 319: Is the comment deprecated now? If so, remove it?

pmbittner (changes requested):

pmbittner:
Changes and design decisions look good to me! Only two minor comments.

@AlexanderSchultheiss Do you want to have a look? Do these changes please your use case? If you are short on time it is sufficient to read the documentation by @ibbem at the top of this PR.

AlexanderSchultheiss:
The changes look good to me! Thank you for your time on this!

src/main/java/org/variantsync/diffdetective/variation/tree/VariationNode.java
ibbem: Good catch. That a mistake.

src/main/java/org/variantsync/diffdetective/variation/diff/parse/DiffTreeParser.java
ibbem: No this comment is still correct (although in very unlikely use cases). Consider
```c
# /* No does something like this, right? */ \
 ifdef A
// This is legal C and C++
#endif
```

PR #83, opened 2023-03-17
Minor fixes and features

pmbittner:


src/main/java/org/variantsync/diffdetective/diff/text/DiffLineNumberRange.java
```diff
@@ -1,4 +1,18 @@
 package org.variantsync.diffdetective.diff.text;
 
+import java.util.Objects;
+
 public record DiffLineNumberRange(DiffLineNumber from, DiffLineNumber to) {
+    @Override
+    public boolean equals(Object o) {
```
ibbem, line 7: All records have generated `equals` and `hashCode` (and `toString`) methods to reduce boilerplate (see [The Java® Language Specification](https://docs.oracle.com/javase/specs/jls/se17/html/jls-8.html#jls-8.10.3)). There is no need to implement them yourself.

src/main/java/org/variantsync/diffdetective/variation/tree/VariationTree.java
```diff
@@ -121,6 +125,13 @@ public static VariationTree fromFile(
         return new VariationTree(tree, source);
     }
 
+    public static VariationTree fromProjection(final Projection projection, final VariationTreeSource source) {
```
ibbem, line 128: This method can be generalized to receive a `VariationNode`.

src/main/java/org/variantsync/diffdetective/variation/tree/VariationTreeNode.java
```diff
@@ -308,6 +308,27 @@ public static VariationTreeNode fromID(int id, List<String> label) {
         );
     }
 
+    public VariationTreeNode deepCopy(final Map<VariationTreeNode, VariationTreeNode> oldToNew) {
```
ibbem, line 311: Document the purpose of oldToNew. Couldn't you just generalize `VariationNode.toVariationTree`? A `deepCopy` alias would still make sense, though.

src/main/java/org/variantsync/diffdetective/variation/tree/VariationTree.java
```diff
@@ -142,6 +153,39 @@ public VariationTree forAllPreorder(final Consumer<VariationTreeNode> action) {
...
+     * The condition might not be invoked on every node in case a node is found.
+     * @param condition A condition to check on each node.
+     * @return True iff the given condition returns true for at least one node in this tree.
+     */
+    public boolean anyMatch(final Predicate<VariationTreeNode> condition) {
+        return anyMatch(root(), condition);
+    }
+
+    /**
+     * Returns tree iff this tree contains the given node.
```
ibbem, line 181: Typo: `tree` -> `true`

src/main/java/org/variantsync/diffdetective/variation/tree/VariationTree.java
```diff
@@ -152,6 +196,14 @@ public int computeSize() {
         return size.get();
     }
 
+    public VariationTree deepCopy() {
+        return deepCopy(new HashMap<>());
+    }
+
+    public VariationTree deepCopy(final Map<VariationTreeNode, VariationTreeNode> oldToNew) {
```
ibbem, line 203: Document the purpose of `oldToNew`.

ibbem (changes requested):
Looks mostly good although some changes are unnecessary or can be generalized.

src/main/java/org/variantsync/diffdetective/variation/tree/VariationTreeNode.java
pmbittner: Is generalized now. Generalization required an abstract method `shallowCopy` to create new nodes which is dependent on subclass.

src/main/java/org/variantsync/diffdetective/variation/tree/VariationTree.java
pmbittner: You are right. Would it make sense though to generalize it? I am unsure what it would mean to create a projection from a variation node. That would probably be up to the variation node itself. Maybe we should leave that for a potential future refactoring? For now having a projection as argument here is more close to the paper and might serve as guidance.

src/main/java/org/variantsync/diffdetective/util/LineRange.java
```diff
@@ -72,4 +77,9 @@ public int getToExclusive() {
     public String toString() {
         return "[" + fromInclusive + ", " + toExclusive + ")";
     }
+
+    @Override
+    public LineRange clone() {
+        return new LineRange(this);
```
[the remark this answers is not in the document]
pmbittner: I refactored LineRange to be a record to emphasize immutability.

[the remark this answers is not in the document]
pmbittner: So we should potentially revert turning LineRange into a record for now or perform the refactoring of hash code methods right here? Would that break something else?

src/main/java/org/variantsync/diffdetective/variation/tree/VariationTree.java
pmbittner: Will be refactored by @ibbem as discussed in meeting.

src/main/java/org/variantsync/diffdetective/util/LineRange.java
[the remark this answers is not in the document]
ibbem: I agree. I'll open an issue for that

src/main/java/org/variantsync/diffdetective/variation/tree/VariationTree.java
ibbem: I meant more of a copy, but I agree the name `fromProjection` is a special case.

src/main/java/org/variantsync/diffdetective/variation/tree/VariationTreeNode.java
ibbem: I though of something different and changed to my liking :)

PR #84, opened 2023-03-17
Fix rename handling

AlexanderSchultheiss:
Changes to file modifiers or file names were expressed in the diff but not part of hunks. If the file content was not changed, this would lead to the meta-data being parsed into the diff tree. 

__Example diffs that caused a problem before the fix__

Mode change:
```diff
diff --git a/main.c b/main.c
old mode 100644
new mode 100755
```

Rename:
```diff
diff --git a/README.md b/README_BAK.md
similarity index 100%
rename from README.md
rename to README_BAK.md
```
In both cases, the lines in the diff would be parsed as artifact lines with mapping `true`. 

__Fix__
If no content changes are found, the `GitDiffer` will now check whether the general git header `diff --git ...` is found. 

If the header is found, the lines in the diff are not taken into account when parsing of the `DiffTree`. Lines in the file are still added to the tree, if the parse options state that non-changed lines should be included. 

If the header is not found, the diff is assumed to stem from some other source (e.g., manually created for tests, copy-pasted from GitHub), and the lines in it are fully processed as before.


pmbittner:
Rebased onto `parse-initial-commit`.

src/test/java/GitDifferTest.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    @ParameterizedTest
+    @MethodSource("testCommits")
+    public void test(String basename) throws IOException, DiffParseException {
+        testCase(basename);
+    }
+
+    private static Repository repo() {
+        final Path marlinPath = Path.of(".")
```
pmbittner, line 45: Name seems to be a copy and paste error. Please rename because there is nothing about marlin here.

src/test/java/GitDifferTest.java
```diff
@@ -0,0 +1,88 @@
...
+            DiffTree diffTree = patch.getDiffTree();
+
+            try (var output = IO.newBufferedOutputStream(actualPath)) {
+                new LineGraphExporter(new Format(new FullNodeFormat(), new ChildOrderEdgeFormat()))
+                        .exportDiffTree(diffTree, output);
+            }
+
+            try (
+                    var expectedFile = Files.newBufferedReader(expectedPath);
+                    var actualFile = Files.newBufferedReader(actualPath);
```
pmbittner, line 75: It seems as if you are writing the actual file to disk and then read it again. I guess it should be possible to keep it in memory and then pass the respective buffer to `IOUtils.contentEqualsIgnoreEOL`?

pmbittner (changes requested):

pmbittner:
Changes look good to me! Only some comments on the test case. @ibbem What is your opinion?

src/test/java/GitDifferTest.java
AlexanderSchultheiss: done

src/test/java/GitDifferTest.java
```diff
@@ -0,0 +1,88 @@
...
+    private static final String[] commitIDs = new String[] {"a032346092e47048becb36a7cb183b4739547370", "4a4c1db1b192e221d8e25460d6d1128d1bdd0c0d", "d7eeba1b94f3d0c0ebed3457eea8fa8537143348"};
+
+    public static Stream<String> testCommits() throws IOException {
+        return Arrays.stream(commitIDs);
+    }
+
+    @ParameterizedTest
+    @MethodSource("testCommits")
+    public void test(String basename) throws IOException, DiffParseException {
+        testCase(basename);
```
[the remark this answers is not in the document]
AlexanderSchultheiss: I simplified it

src/test/java/GitDifferTest.java
pmbittner: You can leave as is for now. We can still refactor tests later.

pmbittner (approved):

src/test/java/GitDifferTest.java
```diff
@@ -0,0 +1,85 @@
...
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Arrays;
+import java.util.stream.Stream;
+
+import static org.junit.jupiter.api.Assertions.fail;
+
+public class GitDifferTest {
+    private final static Path testDir = Constants.RESOURCE_DIR.resolve("diffs").resolve("differ");
+    private final static Path REPOS_DIR = Path.of("src").resolve("test").resolve("resources").resolve("repos");
```
ibbem, line 30: Please also use `Constants.RESOURCE_DIR` here.

src/test/java/GitDifferTest.java
```diff
@@ -0,0 +1,85 @@
...
+                    // Delete output files if the test succeeded
+                    Files.delete(actualPath);
+                }
+            }
+        }
+    }
+
+    private static Repository repo() {
+        final Path repoPath = Path.of(".")
+                .resolve(REPOS_DIR)
```
ibbem, line 74: `REPOS_DIR` already is an absolute path thus you can use `REPOS_DIR.resolve("test-spl.zip");`.

ibbem (changes requested):
Looks good to me. Just some minor things about handling paths.

src/test/java/GitDifferTest.java
ibbem: AlexanderSchultheiss's assumption is correct. That was the original intention. I will look into only writing the `*_actual*` files to disk if the test fails at some point.

ibbem:
Rebased onto develop and fixed the two comments I had.

PR #86, opened 2023-03-21
fix: no longer parse diff meta-line regarding a newline character at the end of a file

AlexanderSchultheiss:
Diffs may contain a meta-line that states whether the last line in a file ends in a `newline` character. Diff Detective interprets these meta-lines as artifact line and adds them as unchanged artifact node to the `DiffTree`.  

__Example diff:__

```diff
diff --git a/dev.txt b/dev.txt
new file mode 100644
index 0000000..a940712
--- /dev/null
+++ b/dev.txt
@@ -0,0 +1,3 @@
+This is a file on a separate branch
+
+Let us see what happens
\ No newline at end of file
```

__Example diff tree:__
image

With this fix, these meta-lines are ignored and no longer parsed as artifact nodes. Thereby, we also discard the information about the `newline` character, which we might to track explicitly in future versions.

### DIFF
--- a/src/main/java/org/variantsync/diffdetective/diff/git/GitDiffer.java
+++ b/src/main/java/org/variantsync/diffdetective/diff/git/GitDiffer.java
@@ -54,7 +54,8 @@ public class GitDiffer {
     private static final Pattern DIFF_HUNK_PATTERN = Pattern.compile( "^@@\\s-(\\d+).*\\+(\\d+).*@@$");
     private static final Pattern GIT_HEADER_PATTERN = Pattern.compile( "^diff --git .*$", Pattern.MULTILINE);
     private static final Pattern DIFF_HEADER_PATTERN = Pattern.compile( "^\\+\\+\\+.*$", Pattern.MULTILINE);
-    private static final String NO_NEW_LINE = "\\ No newline at end of file";
+    private static final Pattern NO_NEWLINE_PATTERN = Pattern.compile(
+            "(" + StringUtils.LINEBREAK_REGEX.pattern() + ")(?m)\\\\ No newline at end of file$");
 
     private final Git git;
     private final DiffFilter diffFilter;
@@ -343,7 +344,7 @@ public class GitDiffer {
 
 
                 try {
-                    final String fullDiff = switch (diffEntry.getChangeType()) {
+                    String fullDiff = switch (diffEntry.getChangeType()) {
                         case ADD, DELETE -> {
                             if (strippedDiff.isEmpty()) {
                                 // Addition or deletion of an empty file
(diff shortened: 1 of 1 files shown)


pmbittner (approved):

AlexanderSchultheiss:
I found that there was still a bug after my previous fix. In this bug, `no newline` meta-lines were already handled by replacing the meta-line with an empty line in the case of modified files. This lead to inconsistent behavior for added/removed vs. unchanged files. 

However, by having a closer look at the bug, I was able to find a more elegant solution to dealing with `no newline`. I reverted my old changes and replaced them with this solution. In the new version, the `no newline` meta-line is removed without replacement. If the meta-line does not exist, a `newline` is added to the last line in the diff.

src/main/java/org/variantsync/diffdetective/diff/git/GitDiffer.java
```diff
@@ -360,6 +363,17 @@ private static CommitDiffResult getPatchDiffs(
...
 
+                    // Iff a file does not end with a newline character, git adds a meta-line to the diff, which states
+                    // the absence of the newline. If this is the case, we remove the meta-line, in order not to
+                    // parse it as artifact line. If the meta-line does not exist, we add the newline character.
+                    // TODO: In future versions, we might want to track the newline more explicitly
+                    final Matcher newlineMatcher = NO_NEWLINE_PATTERN.matcher(fullDiff);
+                    if (newlineMatcher.find()) {
+                        fullDiff = fullDiff.substring(0, newlineMatcher.start());
+                    } else {
+                        fullDiff += StringUtils.LINEBREAK;
```
pmbittner, line 374: This isn't redundant? I though in this case the file ends with a new line character anyway.

ibbem:
There is a bug: If the last line (without newline) changed, the diff looks like this:
```diff
diff --git a/local-maven-repo/org/variantsync/functjonal/1.0-SNAPSHOT/maven-metadata.xml.sha1 b/local-maven-repo/org/variantsync/functjonal/1.0-SNAPSHOT/maven-metadata.x>
index 7a4cbc7f..5220a677 100644
--- a/local-maven-repo/org/variantsync/functjonal/1.0-SNAPSHOT/maven-metadata.xml.sha1
+++ b/local-maven-repo/org/variantsync/functjonal/1.0-SNAPSHOT/maven-metadata.xml.sha1
@@ -1 +1 @@
-c69b8761ddb78427e4ec86904dbb4ddaf354b24a
\ No newline at end of file
+feb03041c908b3f11611fb2380e6bb9424ff2e63
\ No newline at end of file
```
I think the fix is to use a pattern like `StringUtils.LINEBREAK_REGEX.pattern() + "(?m)\\\\ No newline at end of file$"` and just replace all occurrences by the empty string.

AlexanderSchultheiss:
@ibbem good catch! I did not know that the diff looks like this in that case. I assumed that there is only one meta-line.

src/main/java/org/variantsync/diffdetective/diff/git/GitDiffer.java
AlexanderSchultheiss: Hmm... it did not appear in the tree (as seen in the Debugger), and my analysis did not pick it up. I will have a closer look to see what is going on here.

pmbittner: We should just make sure to not introduce an extra empty line which will appear as a new DiffNode again.

pmbittner: Thank you! :) Would you like to add this finding to your documentation?

PR #93, opened 2023-06-21
Generalize variation trees to support nested variation trees

ibbem:
These commits allow for variation trees whose variants are also variation trees (e.g. the variants of `VariationTree<VariationLabel<LinesLabel>>` are `VariationTree<LinesLabel>`). Currently, there are no methods to create such nested variation trees. These will follow in a separate pull request.


pmbittner (changes requested):
Please document the new label classes. Rest looks awesome!

Issue #97, opened 2023-08-30
Formula extraction: Multi-line C-style comments are not filtered


AlexanderSchultheiss:
# Description
The [CPPDiffLineFormulaExtractor](https://github.com/VariantSync/DiffDetective/blob/f104bdbba88a53f8c87bfcb74535b235410c34e1/src/main/java/org/variantsync/diffdetective/feature/CPPDiffLineFormulaExtractor.java#L21) filters comments in lines containing C preprocessor statements. Currently, only comments are filtered that are fully included in the line. Multi-line comments are not filtered. 

This is a bug because it leads to invalid formulas. 

# Example
The extractor determines the invalid formula `A&&BMulti-lineComment` for the code snippet
```c
#if A && B /* Multi-line
                Comment
                */
```
The expected result is `A&&B`.

# Fix
Multi-line comments need to be filtered after inline comments have been filtered.

AlexanderSchultheiss:
I fixed the bug in 5f6f38d.

pmbittner:
Thanks for reporting the bug and fixing it. The changes look good. Would you like to open a pull-request (targeting the `develop` branch)?

Issue #98, opened 2023-09-14
Formula extraction: Nested parenthesis are not resolved correctly


AlexanderSchultheiss:
# Description
I found instances of `#if` macros that use nested parentheses. Such instances are currently not always extracted correctly, because the regex removes the leading and trailing parentheses. 

# Example 
The formula `#if(STDC == 1) && (defined(LARGE) || defined(COMPACT))` is extracted as `STDC__EQ__1)&&(DEFINED_LARGE||DEFINED_COMPACT`. 

The expected correct result is `__LB__STDC__EQ__1__RB__&&(DEFINED_LARGE||DEFINED_COMPACT)`. Note: Parenthesis belonging to boolean expression are abstracted as `__LB__` and `__RB__`.

AlexanderSchultheiss:
I fixed this issue and added a simple test case in [c401591](https://github.com/VariantSync/DiffDetective/commit/c40159184fe3944d0f9dcac37e1ac55247bea94b).

AlexanderSchultheiss:
I found another problematic case that remains unresolved. 
There may be nested formulas that combine logic which we do not abstract with arithmetic expressions that we want to abstract. One example taken from emacs is 
```c
# if ((GNUTLS_VERSION_MAJOR + (GNUTLS_VERSION_MINOR > 0 || GNUTLS_VERSION_PATCH >= 20)) > 3)
```
which currently is extracted as 
```c
((GNUTLS_VERSION_MAJOR__ADD__(GNUTLS_VERSION_MINOR__GT__0||GNUTLS_VERSION_PATCH__GEQ__20))__GT__3)
```

The extraction of the inner expression `GNUTLS_VERSION_MINOR > 0 || GNUTLS_VERSION_PATCH >= 20` is correct, because it follows our current scheme. For example, `#if A >= B && C > D"` should be extracted as `A__GEQ__B&&C__GT__D`, because it follows the scheme `ARITHMETIC_EXPR && ARITHMETIC_EXPR`. 

[...]

PR #99, opened 2023-09-14
Fix issues #97, #98, and #102

AlexanderSchultheiss:
See issues for further details.

### DIFF
--- a/src/main/java/org/variantsync/diffdetective/feature/BooleanAbstraction.java
+++ b/src/main/java/org/variantsync/diffdetective/feature/BooleanAbstraction.java
@@ -139,7 +139,7 @@ public class BooleanAbstraction {
 
     private static final Pattern COMMA = Pattern.compile(",");
     private static final String COMMA_REPLACEMENT = "__";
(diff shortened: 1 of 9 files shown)


pmbittner (approved):
Good catches and thank you for the fixes! :) Changes look good to me. For the fix of #97, the regex now has overlapping groups (match whole multiline comment, or match multiline comment begin). Are you sure that the first group always takes precedence over the second?

AlexanderSchultheiss:
> Good catches and thank you for the fixes! :) Changes look good to me. For the fix of #97, the regex now has overlapping groups (match whole multiline comment, or match multiline comment begin). Are you sure that the first group always takes precedence over the second?

It worked in dozens of test runs that I did since I applied the change.

AlexanderSchultheiss:
I found another edge case of #98 and adjusted the fix.

AlexanderSchultheiss:
I converted the pull request back to a draft, because the current fixes are still insufficient.

AlexanderSchultheiss:
These fixes are now addressed by pull request #103 .

PR #107, opened 2023-10-24
Let maven automatically compile the antlr4 grammar

### LINKED ISSUE #106: Generation of ANTLR files requires ANTLR to be installed
We use ANTLR to generate a parser for formulas within C preprocessor macros. To regenerate the grammar, we have a [script](https://github.com/VariantSync/DiffDetective/blob/develop/scripts/generateANTLRClasses.sh) that runs ANTLR as follows: 

https://github.com/VariantSync/DiffDetective/blob/aa468f0199f7521d9684abe959b5401bcada33ed/scripts/generateANTLRClasses.sh#L13

However, this call requires `antlr4` to be installed locally on the developer's machine. This might be tedious on windows. Given that we already have a maven dependency to antlr4, is there a way to replace this call to `antlr4` with a call to Maven which in turn runs its antlr dependency? Alternatively, we could also ship the antlr4 jar file within the DiffDetective repository and run this jar.

ibbem:
Generated files shouldn't be included in the Git repository when possible. Furthermore, this uses the antlr4 maven plugin as intended and eliminates the requirement of installing antlr4 and running the `generateANTLRClasses.sh` script manually.

This fixes #106 because antlr4 doesn't need to be executed manually anymore. Hence, there is no installation of antlr necessary.


pmbittner (approved):

pmbittner:
Our reasoning was that it is probably simpler to setup DiffDetective when the generated files are included, which have to be regenerated only when the grammar changes anyway. However, your proposed solution looks great. It's surprisingly simple. 😅 I did not expect just a single flag to be enough to generate the code from the grammars automatically.

pmbittner:
closes #106

PR #115, opened 2024-01-26
Update Readme

pmbittner:
The README of DiffDetective is horribly plain and outdated. I added setup instructions, and an overview on our paper's and theses. @ibbem Could you add a quick note on the nix setup (as a new heading `### Setup with Nix` below the `Setup with Maven` instructions)?


pmbittner:
Thanks @ibbem for adding nix instructions. :)

PR #131, opened 2024-02-10
Update nixpkgs

ibbem:
This gets rid of the call to the nixpkgs internal `_fromFetchGit` because the `shallow = true` option has been adopted upstream.


pmbittner (approved):
Great! Thanks!

PR #134, opened 2024-02-21
feat: generic parsing interface and parsing of JavaPP annotations

AlexanderSchultheiss:
# Description
The ArgoUML-SPL has been build using the [Java Preprocessor library](https://www.slashdev.ca/javapp/) which mimics the conditional compilation of the C preprocessor. Due to JavaPP using a slightly different Syntax and being hidden behind line comments, we cannot simply apply our current parsing pipeline to ArgoUML. 

This pull request pull up the necessary interfaced of our existing parsing pipeline to allow for different parser implementations. Additionally, it includes a parser for JavaPP. 

# Notes 
I have integrated this version of DiffDetective in VEVOS and I was able to extract a ground truth for ArgoUML. However, ArgoUML only has one commit in which the variability is edited and this commit appears to contain files with missing 'endif' statements, which causes errors to be reported.


AlexanderSchultheiss:
Have to fix Javadocs

AlexanderSchultheiss:
It seems that the nix build breaks because I incremented DiffDetective's version number. How can I update the build files accordingly?

src/main/java/org/variantsync/diffdetective/feature/AnnotationParser.java
```diff
@@ -0,0 +1,24 @@
+package org.variantsync.diffdetective.feature;
+
+import org.prop4j.Node;
+import org.variantsync.diffdetective.error.UnparseableFormulaException;
+
+public interface AnnotationParser {
```
pmbittner, line 6: Maybe add a short documentation for the purpose of this interface.

src/main/java/org/variantsync/diffdetective/feature/DiffLineFormulaExtractor.java
```diff
@@ -0,0 +1,35 @@
+package org.variantsync.diffdetective.feature;
+
+import org.variantsync.diffdetective.error.UnparseableFormulaException;
+
+/**
+ * Extracts the expression from a C preprocessor statement.
```
pmbittner, line 6: Is this still correct? This is also used for JPP now, right?

src/main/java/org/variantsync/diffdetective/datasets/predefined/Marlin.java
```diff
@@ -13,14 +13,16 @@
...
-        new CPPAnnotationParser(
-            PropositionalFormulaParser.Default,
-            new MarlinCPPDiffLineFormulaExtractor()
-        );
+    public static final PreprocessorAnnotationParser ANNOTATION_PARSER =
+            new PreprocessorAnnotationParser(
+                    PreprocessorAnnotationParser.CPP_PATTERN,
+                    PropositionalFormulaParser.Default,
+                    new MarlinCPPDiffLineFormulaExtractor()
+            );
```
pmbittner, line 21: Instantiating a parser like this looks like it leaks implementation details. In particular, to a new user it might not be clear which patterns to pass here? Would it make sense to introduce some smart constructors (i.e., static methods such as `CreateCPPAnnotationParser` and `CreateJPPAnnotationParser`, which fix / curry the `XPP_PATTERN`)? Not too important though.

src/main/java/org/variantsync/diffdetective/feature/PropositionalFormulaParser.java
```diff
@@ -36,6 +37,11 @@ public interface PropositionalFormulaParser {
             node = FixTrueFalse.EliminateTrueAndFalseInplace(node).get();
         }
 
+        if (node == null) {
+//            Logger.warn("Could not parse expression '{}' to feature mapping. Using it as literal.", fmString);
+            node = new Literal(text);
+        }
```
pmbittner, line 43: This looks like an important fix? Was this code moved? If so, where did it came from?

src/main/java/org/variantsync/diffdetective/feature/cpp/CPPDiffLineFormulaExtractor.java
```diff
@@ -0,0 +1,74 @@
...
+    protected String abstractFormula(String formula) {
+        CExpressionLexer lexer = new CExpressionLexer(CharStreams.fromString(formula));
+        CommonTokenStream tokens = new CommonTokenStream(lexer);
+
+        CExpressionParser parser = new CExpressionParser(tokens);
+        parser.addErrorListener(new ParseErrorListener(formula));
+
+        return parser.expression().accept(new ControllingCExpressionVisitor()).toString();
+    }
+}
```
pmbittner, line 74: This was one of the moved files? Maybe you deleted and inserted it in different commits? That could be a reason why git(hub) does not detect a file move.

src/main/java/org/variantsync/diffdetective/feature/jpp/JPPDiffLineFormulaExtractor.java
```diff
@@ -0,0 +1,49 @@
...
+import org.variantsync.diffdetective.feature.AbstractingFormulaExtractor;
+import org.variantsync.diffdetective.feature.ParseErrorListener;
+import org.variantsync.diffdetective.feature.antlr.JPPExpressionLexer;
+import org.variantsync.diffdetective.feature.antlr.JPPExpressionParser;
+
+import java.util.regex.Pattern;
+
+/**
+ * Extracts the expression from a <a href="https://www.slashdev.ca/javapp/">JavaPP (Java PreProcessor)</a> statement .
+ * For example, given the annotation "//#if defined(A) || B()", the extractor would extract "A || B".
```
pmbittner, line 15: Shouldn't the result be `DEFINED_A || B`?

src/main/java/org/variantsync/diffdetective/variation/NodeType.java
```diff
@@ -1,7 +1,8 @@
 package org.variantsync.diffdetective.variation;
 
-import org.variantsync.diffdetective.variation.diff.DiffNode; // For Javadoc
-import org.variantsync.diffdetective.variation.tree.VariationNode; // For Javadoc
+import org.variantsync.diffdetective.feature.AnnotationType;
+import org.variantsync.diffdetective.variation.diff.DiffNode;
+import org.variantsync.diffdetective.variation.tree.VariationNode;
```
pmbittner, line 5: Please keep the comments `// For Javadoc`.

src/main/java/org/variantsync/diffdetective/variation/NodeType.java
```diff
@@ -37,9 +40,11 @@ public boolean isAnnotation() {
 
     /**
      * Creates a NodeType from its value names.
-     * @see Enum#name()
-     * @param name a string that equals the name of one value of this enum (ignoring case)
+     *
+     * @param name a string that equals the name of one value of this enum (ignoring
+     *             case)
```
pmbittner, line 45: This auto formatting does not always look better than before but its okay. :D

src/main/java/org/variantsync/diffdetective/variation/NodeType.java
```diff
@@ -52,7 +57,36 @@ public static NodeType fromName(final String name) {
...
+                return NodeType.ELIF;
+            }
+            case Else -> {
+                return NodeType.ELSE;
+            }
+            case None -> {
+                return NodeType.ARTIFACT;
+            }
+            default -> throw new IllegalArgumentException(annotationType + "has no NodeType counterpart");
+        }
```
pmbittner, line 85: If you want to be fancy, you might want to use a [switch expression](https://docs.oracle.com/en/java/javase/17/language/switch-expressions.html#GUID-BA4F63E3-4823-43C6-A5F3-BAA4A2EF3ADC) but this looks fine as is. :)

src/test/java/CPPParserTest.java
```diff
@@ -1,152 +1,155 @@
...
+                new TestCase("#if MACRO(x=1)", "MACRO___LB__x__ASSIGN__1__RB__"),
+                new TestCase("#if A = 3", "A__ASSIGN__3"),
+                new TestCase("#if ' ' == 32", "__SQUOTE_____SQUOTE____EQ__32"),
+                new TestCase("#if (NAME<<1) > (1<<BITS)", "__LB__NAME__LSHIFT__1__RB____GT____LB__1__LSHIFT__BITS__RB__"),
+                new TestCase("#if #cpu(sparc)", "CPU___LB__sparc__RB__"),
+                new TestCase("#ifdef \\U0001000", "__B_SLASH__U0001000"),
+                new TestCase("#if (defined(NAME) && (NAME >= 199905) && (NAME < 1991011)) ||     (NAME >= 300000) || defined(NAME)", "(DEFINED___LB__NAME__RB__&&(NAME__GEQ__199905)&&(NAME__LT__1991011))||(NAME__GEQ__300000)||DEFINED___LB__NAME__RB__"),
+                new TestCase("#if __has_warning(\"-Wa-warning\"_foo)",
+                        "__HAS_WARNING___LB____QUOTE____SUB__Wa__SUB__warning__QUOTE_____foo__RB__")
         );
```
pmbittner, line 113: Why is this reindentation necessary? Why should we use 8 spaces instead of 4 spaces now? (Next time I change something in IntelliJ, this styling will be reset and we have huge patches again. 😵 )

pmbittner (changes requested):
Hi @AlexanderSchultheiss ,

these changes look great. Thank you! 😃 The documentation is awesome, too.

I have one major comment: Should we really include the heavy reformatting changes? The reformatting of JavaDoc (e.g., to a certain line width limit) look good to me but many files also have their indentation changed extensively. Next time, I perform some changes in IntelliJ, all the formatting will be reverted, yielding large commits again.

Is this some setting in NeoVim? It might be very hard to change the indentation retroactively, so I can live with the changes. Maybe we should add a style configuration (.editorconfig) to DiffDetective or so.

pmbittner:
> It seems that the nix build breaks because I incremented DiffDetective's version number. How can I update the build files accordingly?

pmbittner:
sorry, hit the wrong button and closed by accident 😓

src/main/java/org/variantsync/diffdetective/feature/PropositionalFormulaParser.java
AlexanderSchultheiss: This code was originally part of CPPAnnotationParser and always executed after PropositionalFormulaParser was called. So I moved it inside the PropositionalFormulaParser. There are no other callers affected by this change.

src/main/java/org/variantsync/diffdetective/variation/NodeType.java
AlexanderSchultheiss: What do you mean? It already is a switch expression?

src/test/java/CPPParserTest.java
AlexanderSchultheiss: I have not used nvim because I was worried about my style being applied to reformat the entire codebase 🙈 . I only used IntelliJ which did this autoformatting.

ibbem:
You missed the version bump in seven other files. Please use something like `rg -F "2.1.0"` or `grep -F "2.1.0" -R` to find all version numbers next time. You can also automatically change these places using `rg -F "2.1.0" --files-with-matches | xargs sed -i 's/2\.1\.0/2.2.0/g'` after verifying that the matches are actually all version numbers.

@pmbittner Should I create an issue for making these files independent of the version number such that we have one canonical place for changing the version number (e.g., `pom.xml`).

ibbem:
I am also a NeoVim user. In case you have problems with your setup I might be able to help you.

src/main/java/org/variantsync/diffdetective/feature/cpp/CPPDiffLineFormulaExtractor.java
AlexanderSchultheiss: I think that all package refactorings were in their own commits. So it is a bit strange that the move is not recognized accordingly.

AlexanderSchultheiss:
> You missed the version bump in seven other files. Please use something like `rg -F "2.1.0"` or `grep -F "2.1.0" -R` to find all version numbers next time. You can also automatically change these places using `rg -F "2.1.0" --files-with-matches | xargs sed -i 's/2\.1\.0/2.2.0/g'` after verifying that the matches are actually all version numbers.
> 
> @pmbittner Should I create an issue for making these files independent of the version number such that we have one canonical place for changing the version number (e.g., `pom.xml`).

Having to do this is not ideal. I searched for the version numbers, but I was unsure which ones I can change safely, because I am not familiar with nix and was not sure whether I would break stuff.  I think that having only one place where the version number is specified would be great.

AlexanderSchultheiss:
> I am also a NeoVim user. In case you have problems with your setup I might be able to help you.

 I have not configured jdtls to use Intellij's style for auto-formatting. How have you integrated Intellij's style in your configuration?

src/main/java/org/variantsync/diffdetective/variation/NodeType.java
AlexanderSchultheiss: They are automatically removed by Intellij... I disabled import optimization.

src/main/java/org/variantsync/diffdetective/feature/PropositionalFormulaParser.java
pmbittner: Ok good idea!

src/main/java/org/variantsync/diffdetective/variation/NodeType.java
pmbittner: It is a switch statement. :) Thought, you might be interested in some new fancy Java syntax. You could write

```java
return switch(annotationType) {
    case If -> NodeType.If;
    case Else -> NodeType.Else;
    ...
};
```

src/test/java/CPPParserTest.java
pmbittner: Oh I see. Seems we have different settings then. I guess its time for a common .editorconfig then. I will open an issue soon then. You dont have to revert the indentation stuff. That is too cumbersome I guess.

pmbittner:
> @pmbittner Should I create an issue for making these files independent of the version number such that we have one canonical place for changing the version number (e.g., `pom.xml`).

That is a good idea, given that also I missed this in my review (even though I incremented the version number correctly just two weeks ago).

src/test/java/CPPParserTest.java
AlexanderSchultheiss: Have you tried formatting he file with your Intellij? I am using the default settings, because I only just installed it on my new Laptop. Perhaps something else in the implementation had changed since the last time it was formatted?

ibbem:
Sorry for the late answer, I totally forgot this.

> > I am also a NeoVim user. In case you have problems with your setup I might be able to help you.
> 
> I have not configured jdtls to use Intellij's style for auto-formatting. How have you integrated Intellij's style in your configuration?

No actually not. I will open an issue for creating an editor config or, if possible, start agreeing on a single formatter.

I also use jdtl, however I only format lines that I actually changed: The `vim.lsp.buf.format` action can also act on a visual selection, so I just select the code I edited and only reformat that particular chunk. It is probably possible to integrate this with Git somehow to only reformat changed lines but I haven't done this yet.

pmbittner:
Hi all,

thanks for opening an issue @ibbem . We can proceed to merge this PR for now and deal with indentation and formatting later. Thanks @AlexanderSchultheiss for all the improvements. :)

src/test/java/CPPParserTest.java
pmbittner: I just formatted this block of TestCases in IntelliJ and the same formatting as you have was applied. Interesting...

pmbittner (approved):

PR #137, opened 2024-03-14
Upgrade to Java 17

pmbittner:
Java 16 is no longer supported so we decided to upgrade to Java 17 which still has long-term support. We also update functjonal in the local-maven-repo to be built with Java 17 now. This should fix some compatibility issues with replication packages.

(Sorry @AlexanderSchultheiss : Squashing our commits made me own all of them. 😅 )


AlexanderSchultheiss (approved):
Made a minor change in pom.xml. Changes look good to me.

PR #139, opened 2024-03-21
Use reproducible time stamps for Maven build artifacts

ibbem:
Unfortunately, JAR files store time stamps of the files inside them.
These time stamps depend on the build time of these files hence breaking
bit for bit reproducibility. By setting all time stamps to a predefined
value and removing superfluous build artifacts this issue can be
avoided. Conventionally, UNIX time stamp 1 signifies an intentionally
set time stamp (in contrast to UNIX time stamp 0) which has no meaning
and is easy to recognize.

I noticed this issue today after reading the following blog post: https://fzakaria.com/2021/06/28/java-nix-reproducibility.html
I also verified that this is indeed an issue in DiffDetective and that it's solved by this PR.


pmbittner:
Interesting! Thanks for the fix! 🥳

PR #148, opened 2024-09-26
CI fixes

pmbittner:
- make CI interruptible: So when we push, and there is a CI job running, then interrupt the current CI job and restart it instead of crashing.
- enable manual workflow runs on Github, see:
  - https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/manually-running-a-workflow
  - https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#workflow_dispatch

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -8,6 +8,7 @@ on:
     branches: [ "main", "develop" ]
   pull_request:
     branches:
+  workflow_dispatch:
 
 jobs:
   build:
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -50,3 +50,9 @@ jobs:
     - name: Publish Javadoc to Github Pages
       id: deployment
       uses: actions/deploy-pages@v1
+
+    # Kill in progress deployments because only the newest version is relevant
+    # and concurrent deployments cause CI failures.
+    concurrency:
+      group: ${{ github.workflow }}-${{ github.ref }}
+      cancel-in-progress: true


.github/workflows/maven.yml
```diff
@@ -50,3 +51,8 @@ jobs:
     - name: Publish Javadoc to Github Pages
       id: deployment
       uses: actions/deploy-pages@v1
+
+    # Marks a job as interruptable by a future pipeline run.
+    # So for example, when we push while there is a CI job running,
+    # the current job gets cancelled and replaced by a new one.
+    interruptible: true
```
ibbem, line 58: Are you sure that this is a feature supported by GitHub? I tried to find documentation for `interruptible` but the only documentation I found was from GitLab not GitHub. [I think GitHub's syntax is a little different](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idconcurrency)

ibbem (changes requested):

.github/workflows/maven.yml
pmbittner: Oh good catch! Guess I was too fast on that one. I tried to fix this now, based on [this example](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow-1). I have no idea though whether this is really what we are looking for, and how to properly test that.

ibbem: So, as far as I understand it, the whole CI is now cancelled when a the same branch is changed while running. However, do we want to cancel the whole CI, including the build and test run, or should we let the build finish and only cancel the GitHub Pages deployment?

pmbittner: I have no idea :D If the problem as only the Github Pages deployment, then we could go for just restarting that. Could you fix the open issues here and adapt the CI to the best of your knowledge? :sweat_smile:

ibbem: OK. Now, only the GitHub Pages deployment is cancelled.

ibbem (approved):

pmbittner:
Thank you! :)

PR #151, opened 2024-11-15
Enhance the nix setup

ibbem:
Mostly necessary to reuse the maven repository for TrueDiffDetective (0e7d810ae5826c9866750f87c7dce4140ac37ed6).
The other changes are fixes and possible enhancements I noticed.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -16,9 +16,9 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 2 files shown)


pmbittner (approved):
Thank you for the improvements. I actually have no idea what is going on I trust you that these are improvements. :sweat_smile: What about adding documentation to the nix file to explain the design, motivation, and purpose of the specific phases and commands. At some point there is just a range of shell commands for example with almost no documentation (only in the commit message).

PR #155, opened 2025-02-14
Improve boolean abstraction

### LINKED ISSUE #154: `#if A` vs `#ifdef A` vs. `#if defined A` vs. `#if defined(A)`
Our heuristic for C preprocessor macros uses some (boolean) abstraction to enable fast and reliable parsing with some sacrifices in accuracy. One such sacrifice is that our current parser does distinguish `#ifdef A` from `#if defined(A)` syntactically and semantically. To improve our parser's accuracy, these two cases could be considered equivalent. As documented in the [CPPParserTest](https://github.com/VariantSync/DiffDetective/blob/main/src/test/java/CPPParserTest.java):

1. `#if A` -> Formula: `A`, AnnotationType: `if`
2. `#ifdef A` -> Formula: `A`, AnnotationType: `if`
3. `#if defined A` -> Formula: `DEFINED_A`, AnnotationType: `if`
4. `#if defined(A)` -> Formula: `DEFINED___LB__A__RB__`, AnnotationType: `if`

On the one hand, these all are syntactically different so it also makes sense to have syntactically different abstractions.
This is violated because our parser abstracts both 1 and 2 to `if`. We do not want to distinguish `#if` from `#ifdef` in the AnnotationType because both are conditional annotations. It could make sense to reflect this difference in the formula then? The other cases are fine because they have syntactically distinct abstract representations.

On the other hand, semantically equal annotations should remain semantically equal. This is violated by all of the above. 2, 3, 4 are semantically equalß. Yet 1 and 2 are parsed to a semantically equal representation and 3 and 4 are completely different because they have completely different formulas. Sacrificing semantic equivalence because of insufficient syntactic analysis is exactly the trade-off of our abstraction. That is fine. It should be consistent and documented though. So <ins>either</ins>  2, 3, 4 should parse to a semantically equivalent abstract representation, which is different from the representation of 1, <ins>or</ins> all of the above should parse to different abstract representations for consistency.

**Proposed fix**: Change the parsing for case 2 in the above list to: `#ifdef A` -> Formula: `DEFINED_A`, AnnotationType: `if`. This would fix both of the problems mentioned above.

This is open to discussion.

Tasks:
- [x] Implement proposed fix or a reasonable alternative
- [x] Document design decisions in our JavaDoc at fitting places.

ibbem:
Fixes #154 by abstracting `#ifdef A` the same as `#if defined(A)`.

As discussed in a meeting with @pmbittner about #154, I checked whether FeatureIDE allows arbitrary feature names (i.e., special characters). This is indeed the case, although, if a formula is printed, it might not be parsable by FeatureIDE anymore.

There are some (at least theoretical) issues with the current boolean abstractions. For example, `#if defined(A)` and `#if DEFINED__LB__A__RB__` are abstracted to the same formula. By directly transforming the parse tree into the formula tree, instead of stringifying and reparsing it with FeatureIDE, we have the opportunity to use feature names which include special characters. This allows us to use the sub-expression unmodified (this PR actually normalizes the expressions somewhat by removing all irrelevant whitespace) and hence prevents all such collisions.

I noticed that `DiffLineFormulaExtractor.resolveFeatureMacroFunctions` is currently never called although there is still one user: `MarlinCPPDiffLineFormulaExtractor`. Should we remove the `DiffLineFormulaExtractor` interface or reimplement this feature?


src/test/java/JPPParserTest.java
```diff
@@ -32,37 +36,40 @@ private record TestCase<Input, Expected>(Input input, Expected expected) {
...
+                new JPPParserTest.TestCase<>("//   #if \"str\" >= \"str\"", new Literal("\"str\">=\"str\"")),
                 // expression := operand > operand
-                new JPPParserTest.TestCase<>("//  #if 1.2 > 0", "1__DOT__2__GT__0"),
+                new JPPParserTest.TestCase<>("//  #if 1.2 > 0", new Literal("1.2>0")),
                 // expression := defined(name)
-                new JPPParserTest.TestCase<>("//#if defined(property)", "DEFINED_property"),
+                new JPPParserTest.TestCase<>("//#if defined(property)", new Literal("defined(property)")),
                 // expression := !defined(name)
-                new JPPParserTest.TestCase<>("//#if !defined(property)", "__U_NOT__DEFINED_property"),
+                new JPPParserTest.TestCase<>("//#if !defined(property)", new Literal("defined(property)", false)),
```
pmbittner, line 58: Maybe it would be a good idea to use and extend our `FormulaUtils` (`import static org.variantsync.diffdetective.util.fide.FormulaUtils.*;`) for these tests. That class defines functions for correction negation of formulas as well as some nice wrappers for the constructors so we do not have to write `new` all the time (e.g., `and(...)` instead of `new And(...)`).

pmbittner (approved):
These changes look great! I am very happy that you were able correct the "bugs" while also simplifying the whole parsing so much! :) Good job!

src/test/java/JPPParserTest.java
pmbittner: As an example, you might want to have a look at `ViewTest`. This is inconsistent across all tests so it might better be done in a new PR.

PR #156, opened 2025-02-26
Fix CI cache restore

ibbem:
I identified the feature interaction that caused sporadic CI failures:
1. The GitHub Pages are built using Nix (the `share/github-pages/DiffDetective/` in the `nix-build` output). However, GitHub Pages uses the [`github-metadata` Jekyll plugin](https://github.com/jekyll/github-metadata) which connects to the GitHub API to retrieve some metadata (which is uninteresting for us). This step is skipped in case there is no internet connection.
2. To decrease CI build times, we currently [cache the Nix installation](https://github.com/rikhuijzer/cache-install) including the Nix store after building all dependencies. Due to a bug (the Nix configuration `/etc/nix/nix.conf` is missing if the Nix installation is restored from the cache), the Nix sandbox settings differ between CI runs. In particular, if there is a cache hit, the Nix sandbox is disabled and thus internet connections become possible during all Nix builds.
The combination of these two problems would result in unreproducible CI builds (which we probably would have never found).
Fortunately (in the sense that this problem became noticeable), there is another interaction that masks these unreproducible CI builds: By default, Nix doesn't have TLS certificates installed in the build environment (they shouldn't be needed anyways). When `github-metadata` tries to access the GitHub API, if there was a cache hit, it fails to validate the GitHub server because the TLS certificate is missing and thus aborts the build.

Problem 1 is effectively solved by https://github.com/jekyll/github-metadata/pull/285. We just need to upgrade the `github-metadata` version or wait until `nixpkgs` adopts that version and update nixpkgs. Then we can just set the environment variable `PAGES_DISABLE_NETWORK` to reliable disable network connections even if the Nix sandbox is disabled.
During investigation of problem 2, I noticed that the [`cache-install` action](https://github.com/rikhuijzer/cache-install), which implements the Nix installation caching for us, is kind of broken by design. I would really like to fix this, unfortunately this design is necessary due to limitations (regarding file permissions) of GitHub's cache action. After some experimentation I found a better approach and a more reliable caching mechanism:
1. I replaced `cache-install` with [install-nix-action](https://github.com/cachix/install-nix-action) which just installs Nix in the GitHub Actions environment but is much faster even without using a similar cache. Though, this new installation doesn't cache our dependencies.
2. After building the dependencies, the Nix store is exported into a single file which can then be cached, restored and imported very reliable, keeping all used components well in their specifications (in contrast to `cache-install` needs to integrate deeply into the Nix installation internals).

During testing I found that change 1 decreases our build times from about [4:35 min](https://github.com/VariantSync/DiffDetective/actions/runs/12651865578/job/35253306654) to about [1:39 min](https://github.com/VariantSync/DiffDetective/actions/runs/14129603083/job/39586382491). Hence, the cache of change 2 doesn't offer the same benefit as before. In particular, the build time when creating the cache increases to about [1:52 min](https://github.com/VariantSync/DiffDetective/actions/runs/14053167559/job/39347105614) and decreases with on a cache hit to about [1:19 min](https://github.com/VariantSync/DiffDetective/actions/runs/14053235003/job/39347286789).
@pmbittner Do you think we should [still implement the cache](https://github.com/VariantSync/DiffDetective/pull/156/commits/da11a13c1451f5a52672fe494ac6e20116346865) or just drop change 2 and stick with only using change 1? Note that with our current development pace, the cache is only rarely useful because it's automatically purged by GitHub after one week (that's the reason the CI failures were only sporadic).


pmbittner:
Thank you so much for figuring this hairy bug out and also fixing the respective bug in the jekyll project!

I appreciate leaving `cache-install` behind if it is so broken. If I understood correctly, the new cache you implemented creates a single big text file with all the nix store package / directory names that were created and that are needed for the CI?

Given your convincing arguments for _not_ having a cache (at least not for DiffDetective), I am in favor of reducing complexity and just going with change 1. Maybe leaving a short comment at the place where the caching would needs to be, which points to this PR would be nice.

So this means we would merge your first commit but drop the second one?

ibbem:
> If I understood correctly, the new cache you implemented creates a single big text file with all the nix store package / directory names that were created and that are needed for the CI?

Yes, that's exactly what it does.

> So this means we would merge your first commit but drop the second one?

I deciced to revert the commit instead of dropping it. This keeps this caching mechanism in the commit history for future reference.

I also pushed a fix for problem 1. It applies the patch for `github-metadata` onto the version that we are currently using. As soon as `github-metadata` tags a new release and it lands in nixpkgs, we might consider updating our dependenies to get rid of this patch.

pmbittner:
sounds good to me :)

Issue #153, opened 2025-01-06
Make the exact way to read the git history configurable


pmbittner:
In DiffDetective, we analyze the git histories of projects. However, we cannot freely configure so far which parts of the history we are looking at (which branch for example). We have a [DiffFilter](src/main/java/org/variantsync/diffdetective/diff/git/DiffFilter.java) to filter out certain commits or files (e.g., merge commits and non C/C++ source code files). We cannot configure the `git log` call though, which is deeply buried in some static functions in the GitDiffer class:
[here](https://github.com/VariantSync/DiffDetective/blob/324a6432740b1e229bbdec464f9dd4492834d846/src/main/java/org/variantsync/diffdetective/diff/git/GitDiffer.java#L79) and
[here](https://github.com/VariantSync/DiffDetective/blob/324a6432740b1e229bbdec464f9dd4492834d846/src/main/java/org/variantsync/diffdetective/diff/git/GitDiffer.java#L98). These should be the only places where we call `git log` but I cannot promise.

Tasks:
- [x] Abstract the call to `git log` to make it configurable. Probably an Interface would be good here with a default implementation (same as current implementation). This interface could then be implemented for other configurations of the git-log-call (e.g., show only commits from main branch for example, or show _all_ commits).
- [x] Move the default DiffFilter from [here](https://github.com/VariantSync/DiffDetective/blob/324a6432740b1e229bbdec464f9dd4492834d846/src/main/java/org/variantsync/diffdetective/datasets/DatasetFactory.java#L41-L47) to the DiffFilter class, so that the Default value can actually be found. Here it would also be good to offer different defaults, depending on whether we are working with C projects or with Java projects for example.

New classes should be located within the same Java package as the other classes linked above.

PR #157, opened 2025-03-31
Release 2.3.0

pmbittner:
- [X] #151 
- [x] #153
- [X] #155 
- [X] #156 
- [X] #158 
- [X] #159
- [X] #160
- [X] Update dependencies
- [x] Increase version number


ibbem:
We might also consider updating all of our dependencies.

pmbittner:
Yes that is a good idea! Could you also have a look at #153?

pmbittner:
@ibbem Could you increase the version number and accept this PR for merge? (I do not think it is actually necessary to re-review everything again, since we already did that as part of all the individual PRs.)

ibbem (approved):

PR #161, opened 2025-06-09
Update dependencies

ibbem:
This is part of #157 and fixes https://github.com/VariantSync/DiffDetective/security/dependabot/6.

Apart from the actual dependency updates (all of which are updates to the newest stable releases), I also added cache invalidation for our Maven dependencies when using Nix (idea by [Peter Kolloch](https://blog.eigenvalue.net/nix-rerunning-fixed-output-derivations/)).


pmbittner (approved):
Great! Thanks!

PR #162, opened 2025-06-09
Fix marlin macro interpretation

### LINKED ISSUE #160: Remove or reimplement `DiffLineFormulaExtractor`
During #155, I noticed that `DiffLineFormulaExtractor.resolveFeatureMacroFunctions` is currently never called (the last call was removed in 5056de8f8f04f74666ff86eb0bae8a83b7572be2) although there is still one user: `MarlinCPPDiffLineFormulaExtractor`. Should we remove the `DiffLineFormulaExtractor` interface or reimplement this feature?

ibbem:
From reading the code, it seems like the Marlin macro interpretation code is still injected. However, it doesn't currently have any effect as it is never called.
This PR fixes #160 and cleans up the interfaces surrounding the annotation parser. It also adds support for C23/C++23 style `#elifdef` and `#elifndef` because I noticed those are missing.


src/main/java/org/variantsync/diffdetective/datasets/predefined/Marlin.java
```diff
@@ -12,10 +13,8 @@
  * @author Kevin Jedelhauser, Paul Maximilian Bittner
```
pmbittner: add your name please :)

src/test/java/CPPParserTest.java
```diff
@@ -120,15 +141,30 @@ private static List<TestCase> testCases() {
...
-                new ThrowingTestCase("#error A"),
-                new ThrowingTestCase("#iferror A"),
+                new TestCase("#if ENABLED(A)", var("A")),
+                new TestCase("#if DISABLED(A)", negate(var("A"))),
+                new TestCase("#if ENABLED(FEATURE_A) && DISABLED(FEATURE_B)", and(var("FEATURE_A"), negate(var("FEATURE_B")))),
+                new TestCase("#if ENABLED(A, B)", var("ENABLED(A,B)")),
+                new TestCase("#if OTHER(A, B)", var("OTHER(A,B)")),
+                new TestCase("#if A", var("A"))
+        );
+    }
```
pmbittner, line 164: nice!

pmbittner (approved):
perfect!

pmbittner:
So I guess integrating the Marlin-Parser is still missing right? We should add it to the default dataset loading I guess.

PR #163, opened 2025-06-09
Abstract `git log` calls

### LINKED ISSUE #153: Make the exact way to read the git history configurable

ibbem:
This fixes #153, makes `GitDiffer` stateless and does a few other small refactorings.

### DIFF
--- a/src/main/java/org/variantsync/diffdetective/datasets/Repository.java
+++ b/src/main/java/org/variantsync/diffdetective/datasets/Repository.java
@@ -1,7 +1,6 @@
 package org.variantsync.diffdetective.datasets;
 
 import org.eclipse.jgit.api.Git;
-import org.eclipse.jgit.api.errors.GitAPIException;
 import org.eclipse.jgit.lib.ObjectId;
 import org.eclipse.jgit.revwalk.RevCommit;
 import org.tinylog.Logger;
@@ -45,6 +44,11 @@ public class Repository {
 	private final String repositoryName;
 
 	/**
+	 * A function that extracts the list of commits that are represented by this repository instance.
+	 */
+	private CommitLister commitLister;
+
+	/**
 	 * Filter determining which files and commits to consider for diffs.
 	 */
 	private DiffFilter diffFilter;
@@ -55,7 +59,7 @@ public class Repository {
 	private PatchDiffParseOptions parseOptions;
 
 	private final Lazy<Git> git = Lazy.of(this::load);
-	
+
 	/**
 	 * Creates a repository.
 	 * 
@@ -63,6 +67,7 @@ public class Repository {
 	 * @param localPath The local path where the repository can be found or should be cloned to.
 	 * @param remote The remote url of the repository. May be <code>null</code> if local.
 	 * @param repositoryName Name of the cloned repository (<code>null</code> if local)
+	 * @param commitLister extracts the commits from {@link #getGitRepo()} that should be represented.
 	 * @param parseOptions Omit some debug data to save RAM.
 	 * @param diffFilter Filter determining which files and commits to consider for diffs.
 	 */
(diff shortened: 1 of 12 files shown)


src/main/java/org/variantsync/diffdetective/datasets/Repository.java
```diff
@@ -50,14 +61,45 @@ public class Repository {
...
+	public static final Function<Repository, Iterator<RevCommit>> AllCommits =
+		(Repository repository) -> {
+			try {
+				return repository.getGitRepo().log().all().call().iterator();
+			} catch (GitAPIException | IOException e) {
+				Logger.warn("Could not get log for git repository {}", repository.getRepositoryName());
+				throw new RuntimeException(e);
+			}
+		};
+
```
pmbittner: I suppose extracting this behavior to an interface (can be an `@FunctionalInterface` so we do not have to change much) would be more idiomatic.

src/main/java/org/variantsync/diffdetective/datasets/Repository.java
```diff
@@ -50,14 +61,45 @@ public class Repository {
...
+			try {
+				return repository.getGitRepo().log().call().iterator();
+			} catch (GitAPIException e) {
+				Logger.warn("Could not get log for git repository {}", repository.getRepositoryName());
+				throw new RuntimeException(e);
+			}
+		};
+
+	/**
+	 * List all commits reachable from a branch of the repository.
```
pmbittner: But which branch is it? All branches?

src/main/java/org/variantsync/diffdetective/datasets/Repository.java
```diff
@@ -66,27 +108,38 @@ public Repository(
...
 			final URI remote,
 			final String repositoryName) {
-		this(repoLocation, localPath, remote, repositoryName,
-                PatchDiffParseOptions.Default, DiffFilter.ALLOW_ALL);
+		this(
+			repoLocation,
+			localPath,
+			remote,
+			repositoryName,
+			TraverseHEAD,
```
pmbittner: So TraverseHEAD is also what we used previously as default right? At least I hope so because that is what I reported in our papers lol.

pmbittner (commented):
Thanks @ibbem! I greatly appreciate the additional simplifications and refactorings. Only some small comments. Can we be sure that these changes are indeed a refactoring? Do we have any tests that cover the behavior of the repository traversal?

src/main/java/org/variantsync/diffdetective/datasets/Repository.java
ibbem: Yes, this was and is the default behavior.

PR #165, opened 2025-06-12
Fix the total commit count

### LINKED ISSUE #164: Total number of commits count changed semantics
During work on #163, I noticed that the [`InvocationCounter` in the `Analysis`](https://github.com/VariantSync/DiffDetective/blob/19a85181d1ec9a5de8057bd3d51b44c2a40eeec0/src/main/java/org/variantsync/diffdetective/analysis/Analysis.java#L408) which counts all commits in the repository (even filtered ones) is never used. This is due to 36dc1bb71c7fac2ea9a09bbeea931d3789a81cac which left that artifact behind.

However, the semantics changed slightly:
Previously, it counted the commits before the `DiffFilter`.
Now, it only counts commits after the `DiffFilter`.
We should decide which behavior is actually intended and document it as such.

Note that the documentation contains some improvement potential in this regard:
- Which behavior does this describe?
  https://github.com/VariantSync/DiffDetective/blob/19a85181d1ec9a5de8057bd3d51b44c2a40eeec0/src/main/java/org/variantsync/diffdetective/analysis/StatisticsAnalysis.java#L41
- This formula contains a typo and implies that `FilterAnalysis` filters commits but it only filters `VariationDiff`s.
  https://github.com/VariantSync/DiffDetective/blob/19a85181d1ec9a5de8057bd3d51b44c2a40eeec0/src/main/java/org/variantsync/diffdetective/analysis/StatisticsAnalysis.java#L24

Furthermore, there is a bug in `Analysis.forSingleCommit`: It [increases the total commits](https://github.com/VariantSync/DiffDetective/blob/19a85181d1ec9a5de8057bd3d51b44c2a40eeec0/src/main/java/org/variantsync/diffdetective/analysis/Analysis.java#L330) although [this is already done in `processCommit`](https://github.com/VariantSync/DiffDetective/blob/19a85181d1ec9a5de8057bd3d51b44c2a40eeec0/src/main/java/org/variantsync/diffdetective/analysis/Analysis.java#L563). So the result will have a total commit count of 2. (Note that I didn't test this. This is just my mental model of the code.)

Regarding `FilterAnalysis`: This could in principle replace `DiffFilter` as implied by the comment in `FilterAnalysis`. For commits, this would already work. For patches, we would need to introduce a new hook, something like `beginParsePatch` (because commits are fully parsed before `beginPatch` is ever called), and do a bunch of refactorings.

ibbem:
Fixes #164.

### DIFF
--- a/src/main/java/org/variantsync/diffdetective/analysis/Analysis.java
+++ b/src/main/java/org/variantsync/diffdetective/analysis/Analysis.java
@@ -325,8 +325,8 @@ public class Analysis {
 
         final double runtime = clock.getPassedSeconds();
         Logger.info("<<< done in {}", Clock.printPassedSeconds(runtime));
-        
-        result.get(TotalNumberOfCommitsResult.KEY).value++;
+
+        result.get(TotalNumberOfCommitsResult.KEY).value = 1;
 
         exportMetadata(analysis.getOutputDir(), result);
         return result;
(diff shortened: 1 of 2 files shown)


pmbittner (approved):
I trust you that you tested this. :D I cannot follow the changes here from just looking at them.

Issue #136, opened 2024-03-04
Consistent code formatting


ibbem:
Which code formatter should we use, without forcing developers into using a specific IDE?
 
@pmbittner is using Intellij with its default settings, right?
@AlexanderSchultheiss and I are using jdtls with its default settings.
Some code is committed without going through a code formatter.

Both are quite configurable. It seems like the [settings of Intellij](https://www.jetbrains.com/help/idea/configuring-code-style.html#import-export-schemes) can be exported as [a jdtls settings file](https://github.com/redhat-developer/vscode-java/wiki/Formatter-settings) called Eclipse profile. @pmbittner could you export your formatting settings, then I will check how different Intellij's and jdtls' formatters handle these settings. If there are no differences we can just commit both setting files into the repo and all current developers should be happy.
If they differ too much we should decide on one formatter which can be integrated into the different IDEs.

What do you think about a precommit hook or CI for code formatting to ensure the code is/stays consistent?

PR #167, opened 2025-06-12
Create a `.editorconfig` and apply it to all source files

ibbem:
This prevents annoying issues with inconsistent formatting (most notably it should enforce consistent whitespace usage in new files) and misconfigured editors (e.g., editors which always try to fix the formatting of the whole file, without explicit request). This also starts tackling #136 whose motivation was exactly the same, although mostly limited to Java files.


pmbittner (approved):
Looks great to me! Thanks for the helpful comments in the commit messages, in particular regarding `git blame`.

pmbittner:
I leave it to you when to merge this PR when you like, in case it causes annoying merge conflicts with other parallel changes.

PR #169, opened 2025-08-11
Improved Unparser

ibbem:
This is a rebase of #152 with some additional fixes and refactorings.

During the rebase I applied the following changes on each commit:
- format code
- fix compilation errors
- temporarily disable broken tests
- remove licensed test cases
- fix documentation (only typos and missing parameter annotations)
- translate commit messages
- consistent author name and email (all commits use `eshul <eugen-shulimov@web.de>` instead of `DARIA-ACER\eshul <shulimov@mail.uni-paderborn.de>` which was only used in the beginning)

Except two commits, which became empty, all commits where kept.

The following was addressed after the rebase in separate commits:
- refactor the tests
- fix a bug in the variation diff `endif` parsing (a8065d9d82a78a24493227a10866bacd46cddd35)
  Note that all tests cases, even the ones that where removed during the rebase, pass the test suit after this fix.
- refactor the variation diff parser (0db0fdd7a818202c36ac2ed3ddcdd5ea6aa76a43)
- improve some JavaDoc comments (9c91b089437a84b5137ff034aa78455dafddac2f)
- refactor `VariationUnparser` (fbd6f3c4b62df5fcc6ba62066b153bdc3ca54800 and 68d81312945886987d78fc723166642ebda01963)
- merge duplicated tree unparsing code (9de8a2634bc76df5fa9b3989805f4e6e9e77dc62)
- fix a bug in `Show.baddiff` which I discovered while debugging the next item in this list (cbf504b55ffacf2816059b24a354d669e998a9df)
- fix a bug in `BadVDiff` which I discovered after implementing the next item in this list (c860a88a950cdd12570c590b4aa801224c8a5de8)
- store `endif`s in the label instead of the nodes directly (de10f75e1cfe8336c71bdb1af97a765ae70226ed)

Notably, I didn't touch the experiment except changing the package name and formatting the code.

There are two things that I'm still unsure about:
1. Should we actually test with `ignoreEmptyLines = true`?
   I don't know how relevant this was to the goal of the thesis. Just judging from the code, it doesn't make sense to test. We could probably get rid of all the `removeWhitespace` business and make the tests much tighter for `ignoreEmptyLines = false`.
2. Should we keep `printSourceCode` or `unparseTree`?
   It turns out that @pmbittner already implemented a variation tree unparser for the views paper. Currently, I removed the stack based implementation because the recursive implementation is more intuitive to me. However, should we keep both interfaces or remove one of the two methods?


src/main/java/org/variantsync/diffdetective/variation/VariationUnparser.java
```diff
@@ -0,0 +1,58 @@
...
+     * @param time that the returned state represents
+     * @return the state before or after the diff
+     */
+    public static String undiff(String diff, Time time) {
+        String excludedDiffSymbol = DiffType.thatExistsOnlyAt(time.other()).symbol;
+
+        String result = diff
+            .lines()
+            .filter(line -> !line.startsWith(excludedDiffSymbol))
+            // TODO assumes that all diff symbols are exactly 1 char long
```
pmbittner: How is this assumption handled in our parser? Can we reuse some constants or parsing methods here?

src/main/java/org/variantsync/diffdetective/variation/diff/DiffNode.java
```diff
@@ -338,14 +338,71 @@ public List<DiffNode<L>> removeChildren(Time time) {
...
+        this.diffType = otherDiffType.inverse();
+        this.from = this.from.as(this.diffType);
+        this.to = this.to.as(this.diffType);
+
+        other.addChildren(this.removeChildren(time), time);
+        getParent(time).replaceChild(this, other, time);
+
+        // FIXME? If we allowed access to the backing node,
+        // we could preserve the identity of this projection by moving it to `other`.
+        this.projections[time.ordinal()] = null;
```
pmbittner: Shouldn't we apply this fix right away? (I am not 100% sure about the problem.)

src/main/java/org/variantsync/diffdetective/variation/diff/DiffNode.java
```diff
@@ -719,7 +776,7 @@ public static <T extends VariationNode<T, L1>, L1 extends Label, L2 extends Labe
 
         for (var variationChildNode : variationNode.getChildren()) {
             var diffChildNode = unchanged(convert, variationChildNode);
-            Time.forAll(time -> diffNode.addChild(diffChildNode, time));
+            diffChildNode.getDiffType().forAllTimesOfExistence(time -> diffNode.addChild(diffChildNode, time));
```
pmbittner, line 795: lol. Was this a bug?

src/main/java/org/variantsync/diffdetective/variation/diff/DiffNode.java
```diff
@@ -754,7 +811,7 @@ public DiffNode<L> shallowCopy() {
             getFromLine(),
             getToLine(),
             getFormula(),
-            Cast.unchecked(label.clone())
+            Cast.unchecked(label.getInnerLabel().clone())
```
pmbittner: This looks like a leaking implementation detail.

src/main/java/org/variantsync/diffdetective/variation/diff/DiffType.java
```diff
@@ -1,11 +1,10 @@
 package org.variantsync.diffdetective.variation.diff;
 
 import org.apache.commons.lang3.function.FailableConsumer;
-import org.tinylog.Logger;
+import org.apache.commons.lang3.function.FailableRunnable;
```
pmbittner, line 4: interesting. I think we also have this in Functjonal.

pmbittner (approved):

src/main/java/org/variantsync/diffdetective/variation/VariationUnparser.java
ibbem: Oh, I assumed we would use `String.startsWith` and then drop the identified substring. However, we actually have the [exact same assumtion](https://github.com/VariantSync/DiffDetective/blob/de10f75e1cfe8336c71bdb1af97a765ae70226ed/src/main/java/org/variantsync/diffdetective/variation/diff/parse/VariationDiffParser.java#L150) 😅.
I see two possibilities to remove this implicit assumption:
- Explicitly check the diff symbols with `String.startsWith` and remove that exact prefix. In case there is no such diff symbol, throw an exception. Note that the variation diff parser already throws an exception in case of an invalid diff (except for empty lines which are interpreted as missing the unchanged diff symbol) but this `undiff` function is currently very permissive in that it will accept any arbitrary string even if not a valid diff.
- Refactor [DiffType.symbol](https://github.com/VariantSync/DiffDetective/blob/de10f75e1cfe8336c71bdb1af97a765ae70226ed/src/main/java/org/variantsync/diffdetective/variation/diff/DiffType.java#L19) from a `String` into a `char`.

src/main/java/org/variantsync/diffdetective/variation/diff/DiffNode.java
ibbem: Well, it was an unintended use case that happens to work with this fix. So, yes. In particular, [`Show.baddiff`](https://github.com/VariantSync/DiffDetective/blob/de10f75e1cfe8336c71bdb1af97a765ae70226ed/src/main/java/org/variantsync/diffdetective/show/Show.java#L54) uses this method and broke my assumption of always using `DiffType.NON`, as demonstrated by the [added test](https://github.com/VariantSync/DiffDetective/blob/de10f75e1cfe8336c71bdb1af97a765ae70226ed/src/test/java/ShowTest.java#L23).

src/main/java/org/variantsync/diffdetective/variation/diff/DiffType.java
ibbem: Oh, yes indeed. There is also https://github.com/VariantSync/Functjonal/issues/4 and I still have a commit lying around which addresses this issue. 😅 Should I continue to work on this issue?

src/main/java/org/variantsync/diffdetective/variation/diff/DiffNode.java
ibbem: I would almost argue the contrary position that the fact that `DiffNode` is using [`VariationLabel`](https://github.com/VariantSync/DiffDetective/blob/de10f75e1cfe8336c71bdb1af97a765ae70226ed/src/main/java/org/variantsync/diffdetective/variation/diff/DiffNode.java#L47) is an implementation detail (although it is currently public and I don't know why). However, passing a [`Variationlabel`](https://github.com/VariantSync/DiffDetective/blob/de10f75e1cfe8336c71bdb1af97a765ae70226ed/src/main/java/org/variantsync/diffdetective/variation/VariationLabel.java#L22) to the `DiffNode` constructor which [wraps the passed label](https://github.com/VariantSync/DiffDetective/blob/de10f75e1cfe8336c71bdb1af97a765ae70226ed/src/main/java/org/variantsync/diffdetective/variation/diff/DiffNode.java#L97) into a `VariationLabel` results in double wrapping. Instead of unpacking and repacking, we could, however, add a second constructor that, instead of receiving a label `L` and a `NodeType`, receives a `VariationLabel`.

src/main/java/org/variantsync/diffdetective/variation/diff/DiffNode.java
ibbem: The problem is this:
```java
DiffNode<L> v = ...;
Projection<L> pb = v.projection(Time.BEFORE);
Projection<L> pa = v.projection(Time.AFTER);
DiffNode<L> vSplit = v.split(time.BEFORE);
// We now have:
// `pa.getBackingNode() == v`
// `pb.getBackingNode() == v` <- this is wrong!
// It should be: `pb.getBackingNode() == vSplit`
// Most modifications to `pb` will now throw and `pb` appears as beeing removed from the tree.

// The current solution is to have `vSplit.projection(Time.BEFORE) != pb`
// although we could get away with a valid `pb` with `pb.getBackingNode() == vSplit` and `vSplit.projection(Time.BEFORE) == pb`.
```
I'll add a commit that fixes this.

src/main/java/org/variantsync/diffdetective/variation/VariationUnparser.java
pmbittner: Ok. I don't mind the exact solution chosen here as long as it is consistent between parser and unparser; and favourably in a way that makes it harder or impossible to introduce an inconsistency later on.

src/main/java/org/variantsync/diffdetective/variation/diff/DiffNode.java
pmbittner: Thank you for adding a fix. This is a very subtle and intricate bug, so good catch!

src/main/java/org/variantsync/diffdetective/variation/diff/DiffNode.java
pmbittner: Since this double wrapping is not visible from the types alone and is a rather weird runtime behavior, I prefer your proposed solution to avoid it via an extra constructor.

src/main/java/org/variantsync/diffdetective/variation/diff/DiffType.java
pmbittner: Yes, I forgot about that issue. I like your proposal from that issue and would be happy to see it resolved.

src/main/java/org/variantsync/diffdetective/variation/VariationUnparser.java
ibbem: I added a commit which refactors `DiffType.symbol` from a `String` to a `char`.

src/main/java/org/variantsync/diffdetective/variation/diff/DiffNode.java
ibbem: I added a commit which applies that proposed change.

PR #171, opened 2025-09-08
Fix the children order in line graph and various refactorings

ibbem:
The line graph exporter tries to merge unchanged children. However, when a child was moved, it cannot be classified as an unchaged edge when linearized (same problems as representing moves in line diffs). This resulted in a bug that makes the line graph exporter non-injective. This was fixed in 3a778b52c15e00345cf413c49ce1cb0ca7085097.

All other commits in this PR are just refactorings I had lying around. All of these commits should probably be reviewed independently but I'm too lazy and thus I'm opening only one PR 😅


src/main/java/org/variantsync/diffdetective/variation/diff/construction/GumTreeDiff.java
```diff
@@ -84,7 +92,18 @@ public static <B extends VariationNode<B, L>, L extends Label> DiffNode<L> diffU
         var dst = new VariationTreeAdapter<L>(after);
 
         MappingStore matching = matcher.match(src, dst);
-        Assert.assertTrue(matching.has(src, dst));
+
+        // We assume that the root nodes are matched.
```
pmbittner: You don't just assume that, you enforce it here. Is it safe to do so?

src/main/java/org/variantsync/diffdetective/util/Assert.java
```diff
@@ -86,7 +86,7 @@ public static void assertFalse(boolean condition, String errorMessage) {
     }
 
     /** Throws {@link AssertionError} with {@code errorMessage} as error message. */
-    public static void fail(String errorMessage) {
+    public static <T> T fail(String errorMessage) {
```
pmbittner, line 89: absurd lol

src/main/java/org/variantsync/diffdetective/variation/diff/DiffNode.java
```diff
@@ -51,32 +50,48 @@ public class DiffNode<L extends Label> implements HasNodeType {
...
 
-    /**
-     * The parents {@link DiffNode} before and after the edit.
-     * This array has to be indexed by {@code Time.ordinal()}
-     *
-     * Invariant: Iff {@code getParent(time) != null} then
-     * {@code getParent(time).getChildOrder(time).contains(this)}.
-     */
-    private DiffNode<L>[] parents = Cast.unchecked(Array.newInstance(DiffNode.class, 2));
+    private static class TimeDependent<L extends Label> {
```
pmbittner: This is non-trivial, so please add some documentation.

pmbittner: From the name, I expected this to be a function `Time -> Value`. This function though is realized by having two fields (which is totally fine) in a DiffNode. For consistency with wording in our papers and my phd thesis, maybe `State` would be a better name? Or `ProjectedFields`? Or `AtTime`? I am unsure. I like the idea of the implementation but the naming is not trivial.

src/main/java/org/variantsync/diffdetective/variation/diff/DiffNode.java
```diff
@@ -267,7 +279,7 @@ public void drop(Time time) {
      * Returns -1 if the given node is not a child of this node.
      */
     public int indexOfChild(final DiffNode<L> child, Time time) {
-        return children[time.ordinal()].indexOf(child);
+        return timeDependent(time).children.indexOf(child);
```
pmbittner: Regarding the naming issue, one could ask here: What name would read well? How would be easiest discovered with IntelliSense?

`at(time).children....`?

`children.at(time)...` / `childrenAt(time)...`?

I like the second term but it breaks with the existing abstraction of `TimeDependent` or would require quite some additional boilderplate.

src/main/java/org/variantsync/diffdetective/variation/diff/DiffNode.java
pmbittner: I appreciate this refactoring! Good idea. :)

ibbem:
In addition to addressing your feedback, I also discovered some bugs and added some tests.

There is one issue that I didn't address: The `GumTreeDiff` doesn't set the line numbers in the `DiffLinesLabel` correctly. However, although this is a bug in all current usages, the implementation is generic in the label and thus cannot generically merge/split labels. There are two possible fixes I can imagine:
- Add merge (and split) function arguments similarly as I did in `DiffNode.join` to fix this issue.
- Add merge (and split) functions to the `Label` interface.
What do you think?

pmbittner:
I like the idea of adding merge and split functions to the label interface. Would that break anything? Do both functions have meaning for our current implementations?

pmbittner (approved):

ibbem:
I added such functions in ec82fd49b24161935b2833f88bea9890a8d5438f. I'm not sure about their names though.

Should we write tests for the line numbers in the `DiffLinesLabel`? Currently, there are only tests for the line numbers in `DiffNode`.

pmbittner:
I clicked on the commit hash link in your comment and made some comments on that commit. Unfortunately, my comments do not appear here, so here are links to them:

- [x] https://github.com/VariantSync/DiffDetective/commit/ec82fd49b24161935b2833f88bea9890a8d5438f#r167331384
- [x] https://github.com/VariantSync/DiffDetective/commit/ec82fd49b24161935b2833f88bea9890a8d5438f#r167331046

pmbittner (approved):

pmbittner:
There seems to be a type error making CI fail. After fixing that, feel free to merge.

PR #173, opened 2025-09-25
Some little features

pmbittner:


src/main/java/org/variantsync/diffdetective/variation/tree/view/relevance/Configure.java
```diff
@@ -35,6 +38,39 @@ public Configure(final Node configuration) {
...
+     * As an example, suppose the map contains the following entries:
+     * <pre>
+     *   A ↦ true
+     *   B ↦ false
+     *   C ↦ true
+     * </pre>
+     * then we construct a formula A ∧ (¬ B) ∧ C.
+     */
+    public Configure(final Map<String, Boolean> assignment) {
+        // We use commutativity if ∧ to iterate the map only once instead of twice as shown in the formula above.
```
ibbem: Typo:
> We use commutativity *of* ∧ to iterate [...]

src/main/java/org/variantsync/diffdetective/variation/diff/VariationDiff.java
```diff
@@ -374,6 +377,24 @@ public int count(final Predicate<DiffNode<L>> nodesToCount) {
         return count.get();
     }
 
+
+    /**
+     * Returns all variable names occurring in annotations (i.e., formulas of mapping nodes) in this variation diff.
+     * @return A set of every occuring feature name.
+     */
+    public Set<String> computeAllFeatureNames() {
```
ibbem: Would be great to have a test for this function.

ibbem (approved):

pmbittner:
I added a test for `VariationDiff::computeAllFeatureNames`, reusing some test cases from other tests. I also made the function return a list to document its deterministic behavior in the type system as well.

PR #174, opened 2025-10-13
Fix a crucial bug in the Configure relevance predicate

pmbittner:
This PR fixes a bug reported by @piameier.

The `Configure` relevance predicate for generating views falsely removed ELSE and ELIF branches. The reason was that the optimized implementation to mark nodes as relevant discarded an entire subtree as soon as that subtree's root was irrelevant (i.e., that root conflicted a given partial configuration). However, in such a case, we still have to check and test any ELSE or ELIF branches.

I implemented the first ever unit tests for views on trees and diffs to check that the bug is now fixed. The new test cases checks the implementation of `Configure` against a naive implementation called `ConfigureSpec`.

In the future, it might be great to
- add more test cases
- add specifications for other relevance predicates as well

This PR also contributes the new [EliminateEmptyAlternatives transformer](https://github.com/VariantSync/DiffDetective/pull/174/commits/1939f504bf73c74a6f8081242cfb23690098bba8) and required refactorings.


src/main/java/org/variantsync/diffdetective/variation/diff/transform/VariationTreeTransformer.java
```diff
@@ -0,0 +1,12 @@
...
+
+import org.variantsync.diffdetective.variation.Label;
+import org.variantsync.diffdetective.variation.tree.VariationTree;
+
+/**
+ * Interface that represents inplace transformations of VariationTrees.
+ * A VariationTreeTransformer is intended to alter a given VariationTree.
+ * @author Paul Bittner
+ */
+public interface VariationTreeTransformer<L extends Label> extends Transformer<VariationTree<L>> {
```
ibbem: What is the benefit of having this additional interface instead of just using `Transformer<VariationTree<L>>`? Same with `VariationDiffTransformer<L>`

src/main/java/org/variantsync/diffdetective/variation/tree/view/relevance/Configure.java
```diff
@@ -73,24 +76,68 @@ public Configure(final Map<String, Boolean> assignment) {
...
+            computeViewNodesOfElifChain(v, markRelevant);
+        }
+    }
+
+    /**
+     * @return true if the given node was marked relevant
+     */
+    private <TreeNode extends VariationNode<TreeNode, ?>> boolean computeViewNodesOfElifChain(TreeNode v, Consumer<TreeNode> markRelevant) {
+        final NodeType vt = v.getNodeType();
+        Assert.assertTrue(vt == NodeType.IF || vt == NodeType.ELSE || vt == NodeType.ELIF);
```
ibbem: This can be simplified:
```suggestion
        Assert.assertTrue(v.isAnnotation());
```

ibbem (approved):
Good work. I did not find any real issues.

Unfortunate that there are commits with failing CI in between. However (as discussed in a meeting), changing that would break downstream work that is already based on this. Hence, we can keep it like that.

Kind of unrelated: Do we check anywhere that a `DiffNode` has at most one ELIF/ELSE? I just checked and didn't find anything. I have a change ready but am kind of surprised that we do not check this yet.

src/main/java/org/variantsync/diffdetective/variation/diff/transform/VariationTreeTransformer.java
pmbittner: Actually these two interfaces caused more trouble than they helped. We can probably remove them.

pmbittner:
Thank you for your review. I do not remember whether we check the ELSE/ELIF validity anywhere. If you have a fix, I am happy to take it. I guess that `DiffNode::assertConsistency` would be a good place? I was also a bit surprised seeing that we do not have `VariationNode::assertConsistency`.

src/main/java/org/variantsync/diffdetective/variation/diff/transform/VariationTreeTransformer.java
pmbittner: I removed the interfaces.

ibbem:
> If you have a fix, I am happy to take it.

I'll open a separate PR.

> I was also a bit surprised seeing that we do not have `VariationNode::assertConsistency`.

We do have [`VariationNode::assertConsistency`](https://github.com/VariantSync/DiffDetective/blob/047df572dfd4b8b2e2974305f65d28d9d3f45f74/src/main/java/org/variantsync/diffdetective/variation/tree/VariationNode.java#L510)

ibbem (approved):

ibbem (approved):

Issue #146, opened 2024-07-11
Unified Interface for Differencers


pmbittner:
Currently, accessing different differencers is a bit messy:

```java
VariationTree<DiffLinesLabel> s1 = VariationTree.fromFile(pathBefore);
VariationTree<DiffLinesLabel> s2 = VariationTree.fromFile(pathAfter);

VariationDiff<DiffLinesLabel> diffViaGit, diffViaGumTree, diffViaTrueDiff;
diffViaGit = VariationDiff.fromFiles(pathBefore, pathAfter,
    DiffAlgorithm.SupportedAlgorithm.MYERS,
    VariationDiffParseOptions.Default);
diffViaGumTree  = VariationDiff.fromTrees(s1, s2);
diffViaTrueDiff = TrueDiffDetective.compare(s1, s2);
```

It would be more transparent if we would have unified interfaces for differencing algorithms such as this for example:
```java
public interface TextBasedDifferencer<L extends Label> {
    public VariationDiff<L> diff(String before, String after);
}
```

```java
public interface TreeBasedDifferencer<L extends Label> {
    public VariationDiff<L> diff(VariationTree<L> before, VariationTree<L> after);
}
```

Then methods could rely on these differencers instead of having lots of convencience methods all around the place as of now.

ibbem:
Addressed in #176

PR #176, opened 2025-10-24
Unified differ interface and source tracing improvements

ibbem:
This PR addresses #146.
Along the way I improved the source tracing because it was necessary to correctly set the source of the returned `VariationDiff`s :). During these changes I also did two other small, unrelated fixes.

open questions:
1. Should we use `Differs.lineDiffer(...).diff(...)` where we currently use other convenience methods?
2. If yes, should we remove the other convenience methods (e.g., `VariationDiff.fromFile`)?.


src/main/java/org/variantsync/diffdetective/variation/diff/Differ.java
```diff
@@ -0,0 +1,48 @@
...
+    /**
+     * Create a variation diff by diffing the content of two {@link String}s.
+     * For {@link VariationDiff#getSource() traceability}, both {@link String}s are paired with a {@link Source}.
+     */
+    default VariationDiff<L> diff(String before, String after, Source beforeSource, Source afterSource) throws IOException, DiffParseException {
+        return diff(new StringReader(before), new StringReader(after), beforeSource, afterSource);
+    }
+
+    /**
+     * Create a variation diff by diffing two {@link String}s.
```
pmbittner: Paths?

ibbem: Good catch.

src/main/java/org/variantsync/diffdetective/util/Source.java
```diff
@@ -0,0 +1,272 @@
...
+import org.variantsync.diffdetective.diff.git.PatchDiff; // For Javadoc
+import org.variantsync.diffdetective.variation.diff.ProjectionSource;
+import org.variantsync.diffdetective.variation.diff.VariationDiff; // For Javadoc
+import org.variantsync.functjonal.Cast;
+
+/**
+ * An interface for source traceability.
+ * For testing and debugging, it can be really useful to know where some value originated from. For
+ * example, imagine an error during you analysis of a {@link VariationDiff}. Now you want to know
+ * where it came from. The solution: DiffDetectice tracks the source and transformations of
```
pmbittner: it -> the variation diff

Otherwise, "it" could also refer to the error but that is not the object we are tracing here.

src/main/java/org/variantsync/diffdetective/util/Source.java
```diff
@@ -0,0 +1,272 @@
...
+ * important data structures like {@link VariationDiff#getSource()} and allows you to inspect them
+ * using this interface.
+ * <p>
+ * {@link Source} represents a hierarchy of inputs, processing steps, and results (the distinction
+ * between these types is not directly reflected in the API and only serve as explanation of the
+ * interface).
+ * <ul>
+ * <li>An input (e.g., {@link FileSource} is a {@link Source} that has no {@link getSources
```
pmbittner: closing bracket after `...FileSource}` is missing.

src/main/java/org/variantsync/diffdetective/util/Source.java
```diff
@@ -0,0 +1,272 @@
...
+ * getSourceExplanation short explanation} and a number of {@link getSourceArguments arguments}. A
+ * source may contain additional data (e.g., the diff a variation diff originated from) that is too
+ * big to be printed by the standard formatting functions ({@link shallowExplanation}, {@link
+ * shortExplanation}, {@link fullExplanation}) of this interface. To access such data, or the
+ * arguments of specific subclasses, you can use {@link findFirst} and {@link findAll}.
+ * <p>
+ * For dealing with generic sources, it is recommended to use the static methods of this interface.
+ * They provide a more readable interface (e.g., {@code variationDiff.shallowExplanation()} vs.
+ * {@code Source.shallowExplanation(variationDiff)}) and correctly deal with {@code null} sources
+ * (treated as {@link Unknown}.
```
pmbittner: closing bracket missing at the very end

src/main/java/org/variantsync/diffdetective/util/Source.java
```diff
@@ -0,0 +1,272 @@
...
+ *
+ * @see CompositeSource
+ * @see FileSource
+ */
+public interface Source {
+    /**
+     * Returns a short, one line explanation or identifier of this source.
+     * The result should be suitable in a format like {@code "sourceExplanation(argument1, argument2)"}.
```
pmbittner: I don't know what it means "to be suitable in a format".

src/main/java/org/variantsync/diffdetective/util/Source.java
```diff
@@ -0,0 +1,272 @@
...
+     * @return an empty list by default
+     */
+    default List<Source> getSources() {
+        return Collections.emptyList();
+    }
+
+    /**
+     * Returns a list of arguments required to understand this source.
+     * Each argument in the result list should represent a {@link Object#toString string} without
+     * newlines suitable in a format like {@code "sourceExplanation(argument1, argument2)"}. This
```
pmbittner: .

src/main/java/org/variantsync/diffdetective/variation/diff/view/DiffView.java
```diff
@@ -72,7 +73,14 @@ public static <L extends Label> BiPredicate<Time, Projection<L>> computeWhenNode
     private static <L extends Label> VariationDiff<DiffLinesLabel> naive(final VariationDiff<L> d, final Relevance rho, final String[] projectionViewText) throws IOException, DiffParseException {
         final VariationDiff<DiffLinesLabel> view;
         try {
-            view = JGitDiff.diff(projectionViewText[0], projectionViewText[1], DiffAlgorithm.SupportedAlgorithm.MYERS, Main.VARIATION_DIFF_PARSE_OPTIONS);
+            view = JGitDiff.diff(
+                projectionViewText[0],
+                projectionViewText[1],
+                Source.Unknown,
+                Source.Unknown,
```
pmbittner: Why not document that the text came from projections of a variation diff?

src/main/java/org/variantsync/diffdetective/variation/diff/view/ViewSource.java
```diff
@@ -1,15 +1,31 @@
...
+ * A {@link Source} that remembers that a variation diff represents a view on another variation
+ * diff.
  * @param diff The original variation diff on which the variation diff with this source is a view on.
  * @param relevance The relevance predicate that was used to create the view.
  */
-public record ViewSource<L extends Label>(VariationDiff<L> diff, Relevance relevance) implements VariationDiffSource {
+public record ViewSource<L extends Label>(VariationDiff<L> diff, Relevance relevance) implements Source {
+    @Override
+    public String getSourceExplanation() {
+        return "View";
```
pmbittner: `"view"` would be consistent with the name of that function in the papers where it is called $view_{tree}$ and $view_{diff}$.

src/main/java/org/variantsync/diffdetective/variation/diff/Differ.java
```diff
@@ -0,0 +1,48 @@
...
+import java.nio.file.Path;
+
+import org.variantsync.diffdetective.diff.result.DiffParseException;
+import org.variantsync.diffdetective.util.FileSource;
+import org.variantsync.diffdetective.util.Source;
+import org.variantsync.diffdetective.variation.Label;
+
+/**
+ * A generic interface for creating variation diffs from the state before and the state after an
+ * edit.
```
pmbittner: And the state is considered to be text, not variation trees. So this interface represents a differ that walks the full visual abstract (i.e., the figure in our README on variability-aware differencing) from start to finish.

Maybe a better name than `Differ` would be `VariabilityAwareDiffer`. This would be consistent with the naming in my PhD thesis.

src/main/java/org/variantsync/diffdetective/variation/diff/Differs.java
```diff
@@ -0,0 +1,134 @@
...
+    /**
+     * Returns a differ that {@link VariationDiffParser#createVariationDiff parses} a
+     * {@link VariationDiff} after {@link JGitDiff#textDiff diffing} with
+     * {@link DiffAlgorithm.SupportedAlgorithm one of JGit's differ}.
+     *
+     * @param lineDiffAlgorithm the Git diffing algorithm to use
+     * @param parseOptions options for parsing the {@link VariationDiff}
+     * @return the diffed {@link VariationDiff}
+     */
+    public static LineDiffer lineDiffer(DiffAlgorithm.SupportedAlgorithm lineDiffAlgorithm, VariationDiffParseOptions parseOptions) {
```
pmbittner: I would prefer a name like `JGit` that makes the underlying engine transparent so that users can decide for an algorithm. There might be many more `LineDiffer`s in the future.

src/main/java/org/variantsync/diffdetective/variation/diff/Differs.java
```diff
@@ -0,0 +1,134 @@
...
+                return parseOptions;
+            }
+        };
+    }
+
+    /**
+     * Calls {@link lineDiffer(DiffAlgorithm.SupportedAlgorithm, VariationDiffParseOptions)} with
+     * {@link DiffAlgorithm.SupportedAlgorithm#MYERS} and {@link VariationDiffParseOptions#Default}.
+     */
+    public static LineDiffer lineDiffer() {
```
pmbittner: This could be called JGitMyers?

src/main/java/org/variantsync/diffdetective/variation/diff/Differs.java
```diff
@@ -0,0 +1,134 @@
...
+
+    /**
+     * Returns a differ that {@link VariationTree#fromFile parses} two variation trees and then
+     * {@link GumTreeDiff#diffUsingMatching diffs} these variation trees.
+     *
+     * @param parseOptions options for parsing the {@link VariationTree}s
+     * @param matcher the matcher used for diffing the {@link VariationTree}s
+     * @return the diffed {@link VariationDiff}
+     */
+    public static TreeDiffer<DiffLinesLabel> treeDiffer(VariationDiffParseOptions parseOptions, Matcher matcher) {
```
pmbittner: `GumTree` as a name for this function?

src/main/java/org/variantsync/diffdetective/variation/diff/Differs.java
```diff
@@ -0,0 +1,134 @@
...
+            }
+        };
+    }
+
+    /**
+     * Calls {@link treeDiffer(VariationDiffParseOptions, Matcher)} with {@link
+     * VariationDiffParseOptions#Default} and {@link Matchers#getInstance {@code
+     * Matchers.getInstance().getMatcher()}}.
+     */
+    public static TreeDiffer<DiffLinesLabel> treeDiffer() {
```
pmbittner: .

src/main/java/org/variantsync/diffdetective/variation/diff/LineDiffer.java
```diff
@@ -0,0 +1,62 @@
...
+
+import org.variantsync.diffdetective.diff.result.DiffParseException;
+import org.variantsync.diffdetective.util.CompositeSource;
+import org.variantsync.diffdetective.util.Source;
+import org.variantsync.diffdetective.variation.DiffLinesLabel;
+import org.variantsync.diffdetective.variation.diff.parse.VariationDiffParseOptions;
+import org.variantsync.diffdetective.variation.diff.parse.VariationDiffParser;
+
+/**
+ * A {@link Differ} that first creates a line diff and
```
pmbittner: I would prefer the name `TextBasedDiffer` over `LineDiffer`. At least "text-based differencing" is the term used in research, our papers, and my thesis. What do you think?

pmbittner (changes requested):
I appreciate these changes and I am very happy about the detailed documentation! Good job!

There are a few things regarding documentation and naming things I would like resolve though before I am willing to merge.

src/main/java/org/variantsync/diffdetective/variation/diff/view/DiffView.java
ibbem: This source is overridden right afterwards. I added comments that clarify this.

I also noticed that the `ViewSource` was not set consistently and fixed it.

src/main/java/org/variantsync/diffdetective/variation/diff/view/ViewSource.java
ibbem: You just want it to be lower case, right?

src/main/java/org/variantsync/diffdetective/variation/diff/LineDiffer.java
ibbem: Using `Text` instead of `Line` makes totally sense but I would drop the `Based` part (i.e., `TextDiffer`) to be consistent with `TreeDiffer`.

src/main/java/org/variantsync/diffdetective/variation/diff/Differ.java
ibbem: I added some text about the state being represented as text vs. trees and included the visual abstract into the javadoc. Note that I moved the actual image from the `docs` into the `src` folder in order to be able to reference is from the Javadoc reliably. For easy linking and referencing, I kept a symlink to that file in the `docs` folder.

The name `VariabilityAwareDiffer` does indeed make more sense. For consistency, we should add the `VariabilityAware`prefix to the other diffing interfaces too though.

src/main/java/org/variantsync/diffdetective/util/Source.java
ibbem: I just want to ensure that implementations expect that `getSourceExplanation` is used in formats like `sourceExplanation(argument1, argument2)` and should thus not use too crazy formatting. 😅

docs/variability-aware-differencing.png
pmbittner, line 1: This change breaks the README file on Github, which displays this image at the top.

pmbittner (commented):
Apart from one error affecting the README, I think this is ready to merge. Thank you very much. :)

src/main/java/org/variantsync/diffdetective/util/Source.java
pmbittner: I see. Thank you for revising the explanation. :)

src/main/java/org/variantsync/diffdetective/variation/diff/Differ.java
pmbittner: Sounds good to me! :+1:

docs/variability-aware-differencing.png
pmbittner: I fixed it by changing the path in the README to point to the original file, not the symlink. I also removed the symlink because it is not used elsewhere.

pmbittner (approved):

pmbittner:
I rebased onto develop to merge now.

PR #177, opened 2025-10-26
Assert that there is at most one else in `assertConsistency`

ibbem:
This was originally discussed in https://github.com/VariantSync/DiffDetective/pull/174#issuecomment-3448232443

### DIFF
--- a/src/main/java/org/variantsync/diffdetective/variation/diff/DiffNode.java
+++ b/src/main/java/org/variantsync/diffdetective/variation/diff/DiffNode.java
@@ -733,6 +733,7 @@ public class DiffNode<L extends Label> implements HasNodeType {
      * Checks that the VariationDiff is in a valid state.
      * In particular, this method checks that all edges are well-formed (e.g., edges can be inconsistent because edges are double-linked).
      * This method also checks that a node with exactly one parent was edited, and that a node with exactly two parents was not edited.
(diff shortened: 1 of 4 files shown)


src/main/java/org/variantsync/diffdetective/variation/diff/DiffNode.java
```diff
@@ -747,6 +747,14 @@ public void assertConsistency() {
             });
         }
 
+        // check that there is at most one ELIF/ELSE
+        Time.forAll(time ->
+            Assert.assertTrue(
+                getChildOrder(time).stream().filter(c -> c.isElif() || c.isElse()).count() <= 1,
+                "There is more than one ELIF/ELSE node"
```
pmbittner: Punctuation missing `.` at end of sentence. Same for VariationNode.

pmbittner (approved):
Thank you. :) As a general question: Wouldn't it make sense to reuse VariationNode::assertConsistency on the projections within DIffNode::assertConsistency? Then we could reuse the ELSE/ELIF check instead of implementing it twice.

ibbem:
You are right, it does make sense to reuse the `assertConsistency`. I updated this PR accordingly.

pmbittner (approved):
Looks great, thank you. :)

PR #179, opened 2026-02-03
Variability-aware patching of software product-line variants

piameier:
This is the code used for the practical part of my master's thesis. We contribute a patching algorithm and an experiment evaluating this patcher and comparing it against gnu patch and mpatch.

src/main/java/org/variantsync/diffdetective/util/StringUtils.java : I do not know if the change of the line break affects code somewhere else, but I needed the change for diffing the unparsed trees with gnu diff. 

src/main/java/org/variantsync/diffdetective/variation/diff/transform/EliminateEmptyAlternatives.java : my IDE shows problems with the .getFirst() method so I replaced it with .get(0).


src/main/java/org/variantsync/diffdetective/variation/diff/VariationDiff.java
```diff
@@ -378,16 +378,16 @@ public int count(final Predicate<DiffNode<L>> nodesToCount) {
      * This method is deterministic: It will return the feature names always in the same order, assuming the variation diff is not changed inbetween.
      * @return A set of every occuring feature name.
      */
-    public LinkedHashSet<String> computeAllFeatureNames() {
-        LinkedHashSet<String> features = new LinkedHashSet<>();
+    public LinkedHashSet<Object> computeAllFeatureNames() {
+        LinkedHashSet<Object> features = new LinkedHashSet<>();
```
pmbittner, line 382: Why was it necessary to turn these strings into objects, only to recover that lost information by casting a few lines later?

src/main/java/org/variantsync/diffdetective/variation/diff/patching/Patching.java
```diff
@@ -0,0 +1,629 @@
+package org.variantsync.diffdetective.variation.diff.patching;
```
pmbittner, line 1: What is this class for? It looks like a bag of utilities. Are these utilities useful only for your experiments or for DiffDetective in general?

It would be good to move utlities that belong to only your experiment into that package, and put the other utilities into the correct places within DiffDetective, and to document them.

src/main/java/org/variantsync/diffdetective/util/StringUtils.java
```diff
@@ -6,7 +6,7 @@
 /** A collection of useful utilities related to string processing. */
 public class StringUtils {
     /** An operating system independent line break used in almost all internal strings. */
-    public final static String LINEBREAK = "\r\n";
+    public final static String LINEBREAK = "\n";
```
pmbittner, line 9: I guess this change is sensible. I do not remember why we chose CRLF here in the first place.

src/main/java/org/variantsync/diffdetective/variation/tree/view/relevance/ConfigureWithFullConfig.java
```diff
@@ -0,0 +1,120 @@
...
+
+import org.prop4j.Node;
+import org.variantsync.diffdetective.util.Assert;
+import org.variantsync.diffdetective.util.fide.FixTrueFalse;
+import org.variantsync.diffdetective.variation.NodeType;
+import org.variantsync.diffdetective.variation.tree.VariationNode;
+
+/**
+ * Relevance predicate that generates (partial) variants from variation trees.
+ * This relevance predicate is the implementation of Equation 5 in our SPLC'23 paper.
```
pmbittner, line 16: Is this documentation up to date? To which predicate in your master's thesis does this belong?

src/main/java/org/variantsync/diffdetective/experiments/thesis_pm/Generator.java
```diff
@@ -0,0 +1,511 @@
...
+import org.variantsync.diffdetective.variation.diff.transform.RevertSomeChanges;
+import org.variantsync.diffdetective.variation.diff.view.DiffView;
+import org.variantsync.diffdetective.variation.tree.VariationTree;
+import org.variantsync.diffdetective.variation.tree.view.TreeView;
+import org.variantsync.diffdetective.variation.tree.view.relevance.Configure;
+import org.variantsync.diffdetective.variation.tree.view.relevance.ConfigureWithFullConfig;
+import org.variantsync.functjonal.Pair;
+import org.variantsync.functjonal.Result;
+
+public class Generator {
```
pmbittner, line 48: What does this generator generate? Could you add some documentation? You can also answer to this comment and I add the JavaDoc documentation later on.

src/main/java/org/variantsync/diffdetective/experiments/thesis_pm/Generator.java
```diff
@@ -0,0 +1,511 @@
...
+	final private static String directory = "experiment";
+	final private static String sourceVariant = "sourceVariant";
+	final private static String version1 = "version1";
+	final private static String version2 = "version2";
+	final private static String targetVariant = "targetVariant";
+	final private static String code = "code.txt";
+	final private static String patch = "patch.txt";
+
+	enum Error {
+		FAILED, ERROR
```
pmbittner, line 61: What is the difference between these?

pmbittner (commented):
Hi Pia,

Thank you very much for the pull-request! I will redirect your PR to a new branch and then do some code cleanup there instead of forcing this work onto you now. I still have some comments and questions though where I could need your help. :)

src/main/java/org/variantsync/diffdetective/util/StringUtils.java
ibbem: As the comment states, we introduced `LINEBREAK` instead of using [`System.lineSeparator`](https://docs.oracle.com/en/java/javase/16/docs/api/java.base/java/lang/System.html#lineSeparator()) to ensure that the line endings of our outputs are independent of the operating system. In particular, if I remember correctly, there was an issue with some dependency (something line graph related according to the Git history) that caused issues on Windows. As windows applications typically take issue with parsing text files without carriage returns while Linux applications just treat the carriage return as data, we use '\r\n' for all outputs.

In summary, the biggest potential issue is probably Windows compatibility. As the output of DiffDetective will not match the expectations in that environment. However, I think we don't use this constant very consistently anyways.

pmbittner: Thank you for the detailed analysis. If I remember correctly, some tests that read/write files failed on Windows but not Linux. :thinking: 

Maybe it is safest to just leave LINEBREAK as is, and maybe add a second constant `LINEBREAK_LF` for code parts of @piameier.

src/main/java/org/variantsync/diffdetective/variation/diff/VariationDiff.java
piameier: I remember that we had problems with other representations of true and false than the strings "True"/"False", e. g., there were also 0 and 1 in the examples. But the change seems a bit weird. Probably, we can drop this change.

src/main/java/org/variantsync/diffdetective/variation/diff/patching/Patching.java
piameier: I sourced the utilities out to a utils class in my thesis directory and documented the functions in the patching class

src/main/java/org/variantsync/diffdetective/variation/diff/VariationDiff.java
pmbittner: Ok. To fix the issues with names for true and false, there is also `org.variantsync.diffdetective.util.fide.FixTrueFalse`.

src/main/java/org/variantsync/diffdetective/variation/diff/patching/Patching.java
pmbittner: perfect, thank you :)

